PR #45, opened 2024-02-15
Docker UI build

vincent99:
Build the UI with the node image in docker, and an Acornfile


njhale (approved):
Tested this out and it works great!

lgtm

PR #55, opened 2024-02-20
Add WinGet install steps and automation

sirredbeard:
Pending acceptance of https://github.com/microsoft/winget-pkgs/pull/140253


sirredbeard:
image

sirredbeard:
Note that GITHUB_TOKEN saved in the repository secrets must be a classic PAT with `public_repo` enabled.

ibuildthecloud:
Cool, we will get this CI setup for this.

drpebcak:
`GITHUB_TOKEN` refers to the builtin token generated by github actions - https://docs.github.com/en/actions/security-guides/automatic-token-authentication

If that does not work, then we need to use some other name to reference a manually generated token.

drpebcak:
I think it will work as is though.. so I am going to merge and see what happens.

Issue #24, opened 2024-02-13
 Builtin support for appending to a file (similar to write file).


sangee2004:
Builtin support for appending to a file would be helpful when there is a need to write to the same file with in a loop. 

I tried the following gptscript to get a file with all acorn commands with their help text.
Using `sys.write` in this case resulted in output file having  help for only  1 acorn command which is as expected.

```
tools: listacorncommands, acornhelpandwrite

List all acorn commands. For each acorn commands from this list , get help text and write the help text to a file acornhelp.txt


---
name: acornhelpandwrite
description: get help text for acorn commands and write to file
tools: sys.write, acornhelp
args: command: acorn command
args: file: file name

get help text for acorn commands and write to a file 

---
name: acornhelp
description: get help text for acorn commands
args: command: acorn command

#!/bin/bash

acorn help ${command}

---
name: listacorncommands
description: List all acorn commands

#!/bin/bash

acorn help
```
It would be helpful to have  `sys.append` in this case.

PR #60, opened 2024-02-21
Append sys.append builtin

njhale:
Add a new built-in tool called `sys.append` that appends content to a file instead of truncating it.

Resolves https://github.com/gptscript-ai/gptscript/issues/24

I think this could also be implemented in `sys.write` by adding an extra parameter to indicate that an append is desired. Interested to hear if folks think that's a better approach.

### DIFF
--- a/pkg/builtin/builtin.go
+++ b/pkg/builtin/builtin.go
@@ -37,6 +37,15 @@ var tools = map[string]types.Tool{
 		},
 		BuiltinFunc: SysWrite,
 	},
+	"sys.append": {
+		Parameters: types.Parameters{
+			Description: "Appends the contents to a file",
+			Arguments: types.ObjectSchema(
+				"filename", "The name of the file to append to",
+				"content", "The content to append"),
+		},
+		BuiltinFunc: SysAppend,
+	},
 	"sys.http.get": {
 		Parameters: types.Parameters{
 			Description: "Download the contents of a http or https URL",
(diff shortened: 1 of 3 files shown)


pkg/builtin/builtin.go
```diff
@@ -251,13 +265,48 @@ func SysWrite(ctx context.Context, env []string, input string) (string, error) {
...
+	var params struct {
+		Filename string `json:"filename,omitempty"`
+		Content  string `json:"content,omitempty"`
+	}
+	if err := json.Unmarshal([]byte(input), &params); err != nil {
+		return "", err
+	}
+
+	// Lock the file to prevent concurrent writes from other tool calls.
+	locker.Lock(params.Filename)
```
njhale, line 289: Added this for some extra protection against interleaving with concurrent writes, but not sure if it's 100% necessary. I'll test without the lock and drop it if it's not necessary.

pkg/builtin/builtin.go
```diff
@@ -251,13 +265,48 @@ func SysWrite(ctx context.Context, env []string, input string) (string, error) {
...
+
+	// Attempt to append to the file and close it immediately.
+	// Write is guaranteed to return an error when n != len([]byte(params.Content))
+	n, err := f.Write([]byte(params.Content))
+	if err := errors.Join(err, f.Close()); err != nil {
+		return "", err
+	}
+
+	msg := fmt.Sprintf("Appended %d bytes to file %s", n, params.Filename)
+	log.Debugf(msg)
```
ibuildthecloud, line 305: Why fmt the string first? Not that it matters but by  passing to log.Debugf(x, ...) it will not take the overhead of string formatting when debug logging is disabled.

njhale: Ah, yup. I just copied this over from SysWrite without thinking about it. I'll fix it.

njhale: done

tylerslaton (approved):

PR #86, opened 2024-02-28
docs: reorganize tool docs

g-linville:
- Moved the Internet search doc to the correct folder
- Reordered the docs folder so that the System Tools are first (I think this makes the most sense)


tylerslaton (approved):
Yeah, I agree with you. Thanks!

StrongMonkey (approved):

PR #99, opened 2024-02-29
docs: add story-book.gpt example

tylerslaton:
This generates a illustrated children's book and serves it at localhost:8081.


examples/story-book.gpt
```diff
@@ -0,0 +1,64 @@
+tools: story-writer, story-illustrator, server, mkdir, sys.write, sys.read, sys.download
+description: Writes a children's book and generates illustrations for it.
+args: story: The story to write and illustrate. Can be a prompt or a complete story.
+
+Do the following steps sequentially:
+
+1. Create a folder called story with mkdir.
+2. If ${story} is prompt and not a complete children's story, call story-writer
```
cjellick, line 8:
```suggestion
2. If ${story} is a prompt and not a complete children's story, call story-writer
```

StrongMonkey (approved):

examples/story-book.gpt
```diff
@@ -0,0 +1,64 @@
...
+   the text and illustrations are side-by-side scrolling down with good padding. Make sure
+   the image is not too large and the text is not too small.
+6. Finally, call serve to serve the story as a webpage and print "Serving at http://localhost:8081".
+
+---
+name: story-writer
+description: Writes a story for children
+args: prompt: The prompt to use for the story
+
+Write a story with a tone for children based on ${prompt}.
```
njhale, line 24: Not sure it matters, but I wonder if the following is less ambiguous for the LLM?

```suggestion
Write a story in the style of a children's book based on ${prompt}.
```

docs/README-USECASES.md
```diff
@@ -81,7 +81,8 @@ Here is a GPTScript that summarizes the the code stored under the current direct
 ### Vision
 Here is an example of a web app that leverages GPTScript to recognize ingredients in a photo and suggest a recipe based on them: [recipe-generator](../examples/recipegenerator).
```
njhale, line 82:
```suggestion
### Vision

Here is an example of a web app that leverages GPTScript to recognize ingredients in a photo and suggest a recipe based on them: [recipe-generator](../examples/recipegenerator).
```

examples/story-book.gpt
```diff
@@ -0,0 +1,64 @@
...
+
+---
+name: story-illustrator
+temperature: 0.5
+# assumes you have image-generation.gpt installed
+tools: ../../image-generation/tool.gpt
```
njhale, line 30: This makes an assumption that `image-generation` is cloned to the same directory as `gptscript` is on the user machine. Do we call that out anywhere as something the user needs to do in order to run this?

njhale (changes requested):
This is really cool!

Not sure it's super important for an example like this, but I noticed is that the images of the characters in a story vary wildly between generated images. For example, my main character started out as a rabbit in the first image but ended up an old man in the last (this was not part of the story's narrative). We may want to summarize the character's appearance and include that in the final prompt that gets sent to image generation.

Apart from that, I left a few other comments.

tylerslaton:
Re-requesting review as this has been substantially overhauled to make things more consistent. Closing open comments for now since they are out-of-date.

examples/story-book.gpt
tylerslaton: There is now a README.md for this example that includes an installation step.

docs/README-USECASES.md
tylerslaton: Fixed!

examples/story-book/story-book.gpt
```diff
@@ -0,0 +1,47 @@
...
+generate an illustration. Append any prompt that you have with "in a children's book
+cartoon style". Only return the URL of the illustration.
+
+---
+name: open
+description: Opens a given file
+args: file-path: Path to the file to open
+
+#!/bin/bash
+open examples/story-book/index.html
```
g-linville, line 47: Does the `open` command exist in Linux? I think this might be a macOS-only thing.

g-linville (approved):

examples/story-book/story-book.gpt
tylerslaton: Looks like for Linux it needs to be xdg-open

examples/story-book/README.md
```diff
@@ -0,0 +1,31 @@
...
+```shell
+(
+    cd image-generation
+    make bootstrap
+)
+```
+
+3. Source the `venv` for `image-generation`.
+```
+source image-generation/venv/bin/activate
```
StrongMonkey, line 24: Looks like this should be `venv/bin/activate` since you are already in `image-generation` folder?

StrongMonkey (approved):

examples/story-book/README.md
tylerslaton: That's a good point. The command above this has `()` around it which means that it won't actually change the user's current shell. However, I can just include the source command in that script so there is no separate command.

StrongMonkey: oh, then actually nvm.

examples/story-book/README.md
```diff
@@ -0,0 +1,40 @@
...
+    ```shell
+    git clone https://github.com/gptscript-ai/image-generation
+    ```
+
+2. **Install and bootstrap the `image-generation` tool.**
+
+    Navigate to the `image-generation` directory, bootstrap the tool, activate the virtual environment, and then return to your original directory:
+
+    ```shell
+    cd ../image-generation
```
cjellick, line 23: Where are you assuming I am when I run this command? The root of the gptscript directory? I just did a git clone, presumably in the dir that is the parent of gptscript. After that, `cd ../image-generation` doesn't make sense

I think you should just drop the cd commands and just tell people where they should be.

cjellick: If you told the user to go to the parent dir of gptscript and then combined the clone with this set of steps, that would make sense. They would all work as a single copy-able block

PR #105, opened 2024-03-01
Chore: Add ability to save state message from each gptscript call

StrongMonkey:
Save each gptscript api calls into Tmpdir with certain filename format. 

This is used by @sanjay920 to analyze and train local LLM to be able to do function calling like gpt4 does.

Saving into tmpdir makes sure we don't pollute the user OS with too much files. 

Developer is expected to upload these file to s3 in a timely fashion. Use `gptscript ./script/upload-s3.gpt` to upload to s3.


tylerslaton (approved):

pkg/runner/runner.go
```diff
@@ -115,6 +119,9 @@ func (r *Runner) call(callCtx engine.Context, monitor Monitor, env []string, inp
 				Type:        EventTypeCallFinish,
 				Content:     *result.Result,
 			})
+			if err := recordStateMessage(result.State); err != nil {
```
njhale, line 122: nit: should a failure to write this debug info cause this method to error out?

njhale (approved):

pkg/runner/runner.go
StrongMonkey: Good point. let's make sure this not fail and log a warning.

PR #108, opened 2024-03-01
add example nodejs and image generation example

cloudnautique:


examples/nodejs-imagegen/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+```bash
+git clone https://github.com/gptscript-ai/gptscript.git
+```
+
+1. Navigate to the 'examples/nodejs-imagegen' directory and install dependencies.
+
+Python:
+
+*note: You can use a virtual environment to install the python dependencies, just make sure it is activated before running the server.*
```
njhale, line 26: nit:

```bash
# (Optional) Create and activate a virtual environment before installing dependencies
python -m venv .venv
source .venv/bin/activate
```

njhale (approved):
@cloudnautique Tested this out and it works great! Awesome example!

I left some suggestions for you to consider, but I don't think any of them are blockers.

cjellick (approved):

PR #115, opened 2024-03-05
Fix: exclude bad env variable when invoking command

StrongMonkey:
We shouldn't allow system level environmental variable to be overridden from tool input, as this will break existing shell functionality. For example, @tylerslaton had this example:

```
tools: mkdir

Call mkdir with "foo"

---
name: mkdir
tools: sys.write
args: path: Path to write to

#!/bin/bash

mkdir $path
```

This will overriden `PATH` to the shell because path is defined as an input. However, this will break existing shell functionality since PATH is overridden and it won't be able to find the executable from `PATH` env variable.

### DIFF
--- a/pkg/engine/cmd.go
+++ b/pkg/engine/cmd.go
@@ -16,6 +16,12 @@ import (
 	"github.com/gptscript-ai/gptscript/pkg/version"
 )
 
+var (
+	reservedEnvNameList = map[string]struct{}{
+		"PATH": {},
+	}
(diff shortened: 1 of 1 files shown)


tylerslaton (commented):
Should we log a warning or something when we do this? Right now it just uses the value without changes which could be confusing from a UX perspective.

StrongMonkey:
Darren already addressed it in a different PR.

PR #117, opened 2024-03-05
Check if response has a choice before debug logging to prevent null r…

darkthread:
The debug logging will cause an exception when the response contains no choice. Adding check logic before using response.Choices[0] can prevent errors.


cjellick:
Do you have a usecase that reproduces?

njhale (approved):
lgtm!

tylerslaton (approved):
Thank you for the contribution @darkthread 👍🏻

Issue #104, opened 2024-03-01
azure doesnt work


cjellick:
See this internal slack thread: https://acorn-io.slack.com/archives/C03FU91U4SX/p1709309998579649

And this discord thread: https://discord.com/channels/1204558420984864829/1204558424734437438/1213154452744314890

And this hacky fix: https://github.com/gptscript-ai/gptscript/pull/103

I don't like the hacky fix though. Seems brittle. Also, seems like a user should not need to set the --default-model flag if they are also setting these azure env vars

sirredbeard:
See also https://github.com/gptscript-ai/gptscript/issues/59

keyallis:
Closing in favor of #43 for tracking purposes

PR #78, opened 2024-02-26
enhance: add support for azure deployment name mapping

keyallis:
#43 

Add a configurable env variable to set the azure deployment name to be used in lieu of the model name. Migrate away from using the snapshotted fork of `go-openai`, and add a hack for passing temperature.

Issue #43, opened 2024-02-15
Make default model an env var to play nice with Azure


dozsa:
I'm trying to run this on Azure and in azure the model name is not actually the model name itself, but instead a user defined model deployment name, which will be user specific. Given that model will be used for all calls, it would help to make the default model name an env var, like the other AzureAI variables. I know I can set the model name for each tool, and tested that with echo, but it is cumbersome to do that for every example and tool i run. Also i tried the hackernews example and though i set the model name for each tool, it still tried to do a call with the default model name.

ibuildthecloud:
Yeah, this makes sense. I've never personally tried azure and the current setting that are in there I just optimistically added hoping it would make it work. We will add this too.

keyallis:
Support should be added and verified

PR #119, opened 2024-03-06
fix: generic azure model support

keyallis:
#43 #104 #78 

Add generic model mapping function for Azure use case. This will require the user to input a default model that is supported by Azure OpenAI, but will provide the least hacky solution, while also being future proof.

Just to note here so it's more easily readable for people who don't want to dive into the docs changes to understand the fix, to invoke you should set the default model to something supported by Azure OpenAI. If you are unsure what models are supported, you can find them by configuring the environment variables and running `gptscript --list-models`. Select one and pass it when invoking the script with the `--default-model` argument.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /docs/yarn.lock
 /releases
 /checksums.txt
+/.env*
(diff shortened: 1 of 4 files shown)


cjellick (approved):
works for me. 👍

PR #120, opened 2024-03-06
docs: update story-book readme to have a reset step

tylerslaton:
This clarifies how a user can run the story-book a second time. We opted to not delete the `pages` directory in the script itself with concerns of UX.


sangee2004 (approved):
LGTM

PR #125, opened 2024-03-07
add sentiments analysis

lucj:
gpt script to get sentiments on a string or text/pdf files


examples/sentiments.gpt
```diff
@@ -0,0 +1,39 @@
+tools: text-from-pdf, sys.read
+description: Get the sentiment of the writer
+args: content: any string
+args: file: path towards a file
+
+You are a program designed to assess the sentiments expressed by a writer in a document.
+
+You can be given a content directly via a string or you could be given the path towards a file.
+
+If you are given a content you can directly get the sentiments from the "${content}"
```
cjellick, line 10: Should this be "given a string"? I dont konw...works as is, so maybe leave it

cjellick:
Per our slack convo, I now think we should have this as a reusable tool. Ive created https://github.com/gptscript-ai/sentiment-analysis

Can you move this there following the emerging pattern we have for tools in repos like https://github.com/gptscript-ai/image-generation, https://github.com/gptscript-ai/vision, and https://github.com/gptscript-ai/question-answerer ?

lucj:
@cjellick I've updated this PR adding an example with the sentiment of the tweet as you suggested.
I'll move the sentiments tool in the dedicated repo.

examples/sentiments.gpt
lucj: I'll clean this up :)

cjellick (approved):

PR #151, opened 2024-03-13
chore: upload s3: remove files containing search results

g-linville:


scripts/upload-s3.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/sh
 
+# Remove any files that could contain search results from a search API
+grep -c "URL:" "${TMPDIR}"/gptscript-state* | grep -v ":0" | cut -d ':' -f1 | xargs rm
```
StrongMonkey, line 4: Should we be worried if other response data could have `URL:` that's not actually from search tool?

StrongMonkey (approved):

thedadams (approved):

PR #152, opened 2024-03-13
Fix: Use subPath as root dir when invoking runtime

StrongMonkey:
When reference a tool in a subDir, for example

```
tools: github.com/gptscript-ai/search/duckduckgo, sys.http.html2text?

summarize nba standings for last season. Search using DuckDuckGo. Feel free to get the contents
of the returned URLs in order to get more information. Provide as much detail as you can.
```

The runtime is being called at the root level, not the subPath. Normally the file(main.go) will live in the same directory as tool.gpt, so runtime will fail. It probably makes more sense to assume all the files are living in the same directory within the subDir that it is referencing to.

Also, in go runtime, building in subDir will fail because of missing git VCS info. Adding `-buildvcs=false` should fix that.

### DIFF
--- a/pkg/repos/get.go
+++ b/pkg/repos/get.go
@@ -78,7 +78,7 @@ func (m *Manager) setup(ctx context.Context, runtime Runtime, tool types.Tool, e
 		return "", nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


g-linville (approved):
Thanks Daishan!

PR #163, opened 2024-03-14
Fix: just use parsed url.String() for http.get

StrongMonkey:
We are just joining url and path with `/` and that would cause extra `/` get appended.

We should just remove the url reassignment. 

Related error:
```
gptscript https://get.gptscript.ai/echo.gpt --input 'Hello, World!'
2024/03/14 16:49:21 error loading https://get.gptscript.ai//echo.gpt: 404 Not Found
```


g-linville:
Are we 100% sure that doing this won't break something else?

ibuildthecloud:
I'm 100% sure it will :)

PR #170, opened 2024-03-15
Fix: fix existing examples to be able to use remote tools

StrongMonkey:
This PR fixed some of the examples that can be replaced by tool packaging.

Specifically:
recipegenerator: don't have to clone vision tool and call it directly through github url.
storybook: being able to use image-generation tool through github url.


sangee2004 (approved):
LGTM

PR #181, opened 2024-03-25
bug: tool should be a synonym for tools, not name

ibuildthecloud:


pkg/parser/parser.go
```diff
@@ -92,6 +90,8 @@ func isParam(line string, tool *types.Tool) (_ bool, err error) {
 		tool.Parameters.InternalPrompt = &v
 	case "export":
 		tool.Parameters.Export = append(tool.Parameters.Export, csv(strings.ToLower(value))...)
+	case "tool":
+		fallthrough
 	case "tools":
 		tool.Parameters.Tools = append(tool.Parameters.Tools, csv(strings.ToLower(value))...)
```
njhale, line 96: nit: alternative syntax

```suggestion
	case "tools", "tool":
		tool.Parameters.Tools = append(tool.Parameters.Tools, csv(strings.ToLower(value))...)
```

njhale (approved):

pkg/parser/parser.go
ibuildthecloud: oh yeah, i always forget about that syntax.

Issue #13, opened 2024-02-08
Add colors to terminal output


g-linville:
Right now, everything just prints in the default color. I think it would be nice to have function calls show up in one color, output from the LLM show up in another etc., as this would make it easier to visually scan through the output of a single `gptscript` invocation.

cjellick:
TUI makes this better - it has colors :-)

PR #182, opened 2024-03-25
enhance: add green color to tool call names

g-linville:
related to #13

When scanning through the output of gptscript, I am usually most interested in finding the tool calls. I figured that making the tool names green would make it easier.

We could easily add more colors to the output as well, but I wasn't really sure what else to do.

Here's an example of what it looks like:
image

### DIFF
--- a/pkg/cli/gptscript.go
+++ b/pkg/cli/gptscript.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"github.com/acorn-io/cmd"
+	"github.com/fatih/color"
 	"github.com/gptscript-ai/gptscript/pkg/assemble"
 	"github.com/gptscript-ai/gptscript/pkg/builtin"
 	"github.com/gptscript-ai/gptscript/pkg/cache"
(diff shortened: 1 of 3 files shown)


pkg/cli/gptscript.go
```diff
@@ -34,6 +35,7 @@ type GPTScript struct {
 	CacheOptions
 	OpenAIOptions
 	DisplayOptions
+	Color         *bool  `usage:"Use color in output (default true)" default:"true"`
```
njhale, line 38: nit: can this just be a bool instead of *bool?

njhale (approved):
One nit, otherwise lgtm

pkg/cli/gptscript.go
g-linville: I was following the example of the `--cache` flag, which you can see here: https://github.com/gptscript-ai/gptscript/blob/fd7e77df85168b79788629e9bd1000cab21c2dc5/pkg/cache/cache.go#L21

It uses a bool pointer. I noticed that when I had just a bool, the `default:"true"` struct tag didn't do anything, but it worked as soon as I made it a pointer.

g-linville: Also I just noticed that the description for `--cache` is the opposite of what it's supposed to be, lol

thedadams (approved):

PR #183, opened 2024-03-25
fix: detect built-in tools when loading from assemble output

thedadams:
Using the --assemble flag will output the contents that can be read again by gptscript to construct a tool without going through all the steps again. This doesn't detect built-in tools, and will, instead, try to run the built-in tools like they are commands.

This change will detect the built-in tools to ensure they are run correctly.


thedadams:
Issue: https://github.com/gptscript-ai/gptscript/issues/184

g-linville (approved):

PR #195, opened 2024-03-28
feat: use an OpenAPI definition as a tool file

g-linville:
This adds the ability to use an OpenAPI definition file as though it were a GPTScript tool file. gptscript will now attempt to parse any .json, .yaml, or .yml files as OpenAPI definitions, and if it gets an error while doing that, only then will it attempt to treat it as a standard gptscript file. (Files ending in any other extension are just treated as standard scripts and not checked for OpenAPI definitions.)

When parsing the file, gptscript creates an in-memory tool definition for each operation (path + HTTP method) defined in the file. It also creates one top-level tool that exports all the rest. Each tool's name matches the corresponding operation's `operationId`. Each tool's instructions (the bit after the preamble, where you would ordinarily prompt the model or run a command) starts with the unique sequence `#!sys.openapi`, followed by a JSON string with information about the operation. This tells gptscript's tool runner to handle this tool uniquely. The code that runs the tool will unmarshal the JSON and use the instructions to set up and do the HTTP request, print the body, and exit.

The typical use case for this would be to hand an OpenAPI definition to GPTScript and ask the model to just perform a series of tasks, and let it choose which operations it needs to use. It is still possible to restrict which tools you give it using the `toolName from file.yaml` syntax.

For more details on how to use this, see the doc that I added in this PR.


pkg/types/jsonschema.go
```diff
@@ -3,11 +3,15 @@ package types
...
+	Description string      `json:"description,omitempty"`
+	Type        string      `json:"type,omitempty"`
+	Ref         string      `json:"$ref,omitempty"`
+	Items       *JSONSchema `json:"items,omitempty"`
+	Enum        []string    `json:"enum,omitempty"`
 
 	ID         string                `json:"$id,omitempty"`
 	Title      string                `json:"title,omitempty"`
-	Properties map[string]Property   `json:"properties"`
+	Properties map[string]JSONSchema `json:"properties"`
```
g-linville, line 14: These are some fixes that I made to the JSONSchema. There was no need to have a separate `Property` struct.

pkg/engine/openapi.go
```diff
@@ -0,0 +1,383 @@
...
+	// Set up the request
+	req, err := http.NewRequest(instructions.Method, u.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	// Check for a bearer token (only if using HTTPS)
+	if u.Scheme == "https" {
+		// For "https://example.com" the bearer token env name would be GPTSCRIPT_EXAMPLE_COM_BEARER_TOKEN
+		bearerEnv := "GPTSCRIPT_" + strings.ToUpper(strings.Replace(u.Host, ".", "_", -1)) + "_BEARER_TOKEN"
```
ibuildthecloud, line 75: I think you want u.Hostname() so that you don't get the port part if there is a port, like `:8443`

g-linville: Thanks for finding that!

g-linville: Fixed

ibuildthecloud (approved):

pkg/loader/loader.go
```diff
@@ -126,9 +127,24 @@ func readTool(ctx context.Context, prg *types.Program, base *source, targetToolN
 		return loadProgram(data, prg, targetToolName)
 	}
 
-	tools, err := parser.Parse(bytes.NewReader(data))
-	if err != nil {
-		return types.Tool{}, err
+	var tools []types.Tool
+	if strings.HasSuffix(base.Name, ".json") ||
```
ibuildthecloud, line 131: We should detect if the data is openapi format by inspecting the data itself and not by file extension. That should be more flexible in the future. I also don't want something where the openai schema is hosted at path like "https://example.com/openapi" and it not working becaues it doesn't end with the extension.

pkg/engine/openapi.go
```diff
@@ -0,0 +1,383 @@
...
+	// Set up the request
+	req, err := http.NewRequest(instructions.Method, u.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request: %w", err)
+	}
+
+	// Check for a bearer token (only if using HTTPS)
+	if u.Scheme == "https" {
+		// For "https://example.com" the bearer token env name would be GPTSCRIPT_EXAMPLE_COM_BEARER_TOKEN
+		bearerEnv := "GPTSCRIPT_" + strings.ToUpper(strings.Replace(u.Hostname(), ".", "_", -1)) + "_BEARER_TOKEN"
```
ibuildthecloud, line 75: There's a function somewhere ToEnvLike.  Can you use that.  It's handles `-` -> `_` also which tends to break env vars also.

g-linville: @ibuildthecloud that function doesn't replace `.` with `_`. Should I modify it to do that also?

g-linville: Nvm, I hadn't updated my local branch on main yet. Fixed.

pkg/loader/loader.go
g-linville: Fixed

ibuildthecloud (approved):

PR #207, opened 2024-04-01
enhance: openapi: add support for authentication

g-linville:
This adds support for security schemes in the OpenAPI feature.

### DIFF
--- a/pkg/engine/openapi.go
+++ b/pkg/engine/openapi.go
@@ -88,6 +88,13 @@ func (e *Engine) runOpenAPI(tool types.Tool, input string) (*Return, error) {
 		if err := handleAuths(req, envMap, instructions.SecurityInfos); err != nil {
 			return nil, fmt.Errorf("error setting up authentication: %w", err)
 		}
+
+		// If there is a bearer token set for the whole server, and no Authorization header has been defined, use it.
+		if token, ok := envMap["GPTSCRIPT_"+env.ToEnvLike(u.Hostname())+"_BEARER_TOKEN"]; ok {
+			if req.Header.Get("Authorization") == "" {
+				req.Header.Set("Authorization", "Bearer "+token)
+			}
+		}
 	}
 
 	// Handle query parameters
--- a/pkg/engine/openapi.go
+++ b/pkg/engine/openapi.go
@@ -15,7 +15,10 @@ import (
 	"github.com/tidwall/gjson"
 )
 
-var SupportedMIMETypes = []string{"application/json", "text/plain", "multipart/form-data"}
+var (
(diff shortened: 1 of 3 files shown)


thedadams (approved):

pkg/loader/openapi.go
```diff
@@ -142,14 +165,63 @@ func getOpenAPITools(t *openapi3.T) ([]types.Tool, error) {
...
 			}
 
+			// See if there is any auth defined for this operation
+			var noAuth bool
+			if operation.Security != nil {
+				if len(*operation.Security) == 0 {
+					noAuth = true
+				}
+				for _, req := range *operation.Security {
+					current := map[string]any{}
```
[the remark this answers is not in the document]
g-linville: Oh, thanks. Something felt wrong to me about the way that I implemented this but I couldn't figure out what it was.

pkg/engine/openapi.go
```diff
@@ -70,12 +83,17 @@ func (e *Engine) runOpenAPI(tool types.Tool, input string) (*Return, error) {
...
-		if bearerToken, ok := envMap[bearerEnv]; ok {
-			req.Header.Set("Authorization", "Bearer "+bearerToken)
+	// Check for authentication (only if using HTTPS)
+	if u.Scheme == "https" && len(instructions.SecurityInfos) > 0 {
+		if err := handleAuths(req, envMap, instructions.SecurityInfos); err != nil {
+			return nil, fmt.Errorf("error setting up authentication: %w", err)
+		}
+
+		// If there is a bearer token set for the whole server, and no Authorization header has been defined, use it.
+		if token, ok := envMap["GPTSCRIPT_"+env.ToEnvLike(u.Hostname())+"_BEARER_TOKEN"]; ok {
```
vbehar, line 93: @g-linville hi, I just tried the openapi integration with this new code, and using a global _BEARER_TOKEN env var doesn't work anymore, because this code is not reached if I have a security info set, but without the specific _${name} env var.

for example in my case, my openapi document contains:

```yaml
components:
  securitySchemes:
    bearerAuth:
      bearerFormat: JWT
      scheme: bearer
      type: http
```

so if I don't have the `_BEARERAUTH` env var, I get an error - instead of the code trying to lookup the `_BEARER_TOKEN` env var.

that's not a big issue because I can change the env var on my side, but it's just that it might confuse people...

g-linville: @vbehar Thanks for bringing this to my attention. Yes, this was technically a breaking change, but there has been no release of GPTScript that includes the OpenAPI feature (the only way to use it has been to compile it yourself) so that's why I didn't specifically call it out anywhere.

PR #208, opened 2024-04-01
fix: various fixes for openapi support

g-linville:
These are fixes to various problems that Darren found with OpenAPI support. I annotated this PR to explain a bit about each substantial change.

### DIFF
--- a/pkg/loader/loader.go
+++ b/pkg/loader/loader.go
@@ -131,7 +131,11 @@ func readTool(ctx context.Context, prg *types.Program, base *source, targetToolN
 	var tools []types.Tool
 	if isOpenAPI(data) {
 		if t, err := openapi3.NewLoader().LoadFromData(data); err == nil {
-			tools, err = getOpenAPITools(t)
+			if base.Remote {
+				tools, err = getOpenAPITools(t, base.Location)
+			} else {
+				tools, err = getOpenAPITools(t, "")
+			}
 			if err != nil {
 				return types.Tool{}, fmt.Errorf("error parsing OpenAPI definition: %w", err)
 			}
@@ -256,6 +260,9 @@ func ProgramFromSource(ctx context.Context, content, subToolName string) (types.
 }
 
 func Program(ctx context.Context, name, subToolName string) (types.Program, error) {
+	if subToolName == "" {
+		name, subToolName = SplitToolRef(name)
+	}
 	prg := types.Program{
 		Name:    name,
 		ToolSet: types.ToolSet{},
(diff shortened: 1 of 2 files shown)


pkg/loader/loader.go
```diff
@@ -318,14 +325,14 @@ func SplitToolRef(targetToolName string) (toolName, subTool string) {
 }
 
 func isOpenAPI(data []byte) bool {
-	var version struct {
-		OpenAPI string `json:"openapi" yaml:"openapi"`
+	var fragment struct {
+		Paths map[string]any `json:"paths,omitempty"`
```
g-linville, line 329: We want to check for the existence of the `paths` map, rather than the version string, since the version string is sometimes missing.

pkg/loader/openapi.go
```diff
@@ -15,15 +16,26 @@ import (
...
+		if defaultHost != "" {
+			u, err := url.Parse(defaultHost)
+			if err != nil {
+				return nil, fmt.Errorf("invalid default host URL: %w", err)
+			}
+			u.Path = "/"
+			t.Servers = []*openapi3.Server{{URL: u.String()}}
+		} else {
+			return nil, fmt.Errorf("no servers found in OpenAPI spec")
+		}
```
g-linville, line 31: This adds support for a "default host" if no server is specified at the global level in the OpenAPI document.

pkg/loader/openapi.go
```diff
@@ -80,10 +92,9 @@ func getOpenAPITools(t *openapi3.T) ([]types.Tool, error) {
 				},
 			}
 
-			toolNames = append(toolNames, tool.Parameters.Name)
-
-			// Handle query, path, and header parameters
-			for _, param := range operation.Parameters {
+			// Handle query, path, and header parameters, based on the parameters for this operation
+			// and the parameters for this path.
+			for _, param := range append(operation.Parameters, pathObj.Parameters...) {
```
g-linville, line 97: Parameters can also be defined at the path level (`pathObj.Parameters`), so we are now handling those as well.

pkg/loader/openapi.go
```diff
@@ -136,10 +147,6 @@ func getOpenAPITools(t *openapi3.T) ([]types.Tool, error) {
 					tool.Parameters.Arguments.Properties["requestBodyContent"] = &openapi3.SchemaRef{Value: arg}
 					break
 				}
-
-				if bodyMIME == "" {
-					return nil, fmt.Errorf("no supported MIME types found for request body in operation %s", operation.OperationID)
-				}
```
g-linville, line 142: Rather than failing on the whole document if we encounter a body with no supported MIME types, we just skip it and move on.

pkg/loader/openapi.go
```diff
@@ -138,7 +150,8 @@ func getOpenAPITools(t *openapi3.T) ([]types.Tool, error) {
 				}
 
 				if bodyMIME == "" {
-					return nil, fmt.Errorf("no supported MIME types found for request body in operation %s", operation.OperationID)
+					// No supported MIME types found, so just skip this operation and move on.
+					continue operations
```
g-linville, line 154: We now skip tools with no supported MIME types in the body instead of just returning an error for the whole document.

ibuildthecloud (approved):

thedadams (approved):

PR #225, opened 2024-04-08
Fix: fix golang packaging in windows

StrongMonkey:
This PR fixed some bugs around packaging and invoking golang runtime in windows.

### DIFF
--- a/pkg/engine/cmd.go
+++ b/pkg/engine/cmd.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"runtime"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


pkg/engine/daemon.go
```diff
@@ -172,7 +172,7 @@ func (e *Engine) startDaemon(_ context.Context, tool types.Tool) (string, error)
 		e.Ports.daemonWG.Done()
 	})
 
-	for i := 0; i < 20; i++ {
+	for i := 0; i < 120; i++ {
```
StrongMonkey, line 175: Need to wait for more time to setup remote toolchain(like downloading runtime/pip install) before it can timeout

pkg/engine/cmd.go
```diff
@@ -202,6 +203,17 @@ func (e *Engine) newCommand(ctx context.Context, extraEnv []string, tool types.T
 		cmdArgs = append(cmdArgs, f.Name())
 	}
 
+	// This is a workaround for Windows, where the command interpreter is constructed with unix style paths
+	// It converts unix style paths to windows style paths
```
StrongMonkey, line 207: This only converts the first arg (which is ${GPTSCRIPT_TOOL_DIR}/bin/gptscript-go-tool)`. There could be some edge cases where the rest of the args contain unix path, which is not addressed here.

pkg/engine/daemon.go
StrongMonkey: @ibuildthecloud The problem is that today when we use clicky-chat to download knowledge-api, the clicky-chat daemon will timeout first before it can download the other runtime(python). Ideally this timeout could probably be higher to give more time to download.

PR #229, opened 2024-04-10
Add option to set github auth token

nw0rn:
This PR allows an user to set a Github access/bearer token to authenticate against the Github API. Increases the request limit from [60](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users) to [5000](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-authenticated-users) requests per hour. When playing around with shim providers I reached the limit relatively fast. For me it was not even going away after several hours  - adding an auth token fixed it. 

Avoids this error: 

`HEAD: 403 Forbidden {"message":"API rate limit exceeded for 37.xxx.xxx.xxx. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}`

### DIFF
--- a/pkg/loader/github/github.go
+++ b/pkg/loader/github/github.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


ibuildthecloud:
This is a decent stop gap. The longer term thing we want to do is basically fall back to a pure git approach to this which shouldn't be rate limited.  There's also a hidden feature in gptscript called "assemble" that will kind of compile the script once so it doesn't keep looking it up. But that flow doesn't fully make sense to be and that's why it's hidden. It needs to be rethought.  https://github.com/gptscript-ai/story-book/blob/main/Dockerfile#L5

ibuildthecloud (approved):

PR #239, opened 2024-04-12
fix: credentials: don't store creds for local credential tools

g-linville:
for #240

There was a problem where sometimes credential tools without `github.com` at the front of their name were having their credentials stored anyway. This change also skips the credential lookup for non-GitHub tools, since we aren't going to store the credentials for those anyway.

### DIFF
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -348,7 +348,7 @@ func (r *Runner) handleCredentials(callCtx engine.Context, monitor Monitor, env 
 		)
 
 		// Only try to look up the cred if the tool is on GitHub.
-		if strings.HasPrefix(credToolName, "github.com") {
+		if isGitHubTool(credToolName) {
 			cred, exists, err = store.Get(credToolName)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/runner/runner.go
```diff
@@ -340,9 +341,18 @@ func (r *Runner) handleCredentials(callCtx engine.Context, monitor Monitor, env
...
-		if err != nil {
-			return nil, fmt.Errorf("failed to get credentials for tool %s: %w", credToolName, err)
+		var (
+			cred   *credentials.Credential
+			exists bool
+			err    error
+		)
+
+		// Only try to look up the cred if the tool is on GitHub.
+		if strings.HasPrefix(credToolName, "github.com") {
```
njhale, line 351: nit: I suggest using a constant instead of a literal since we're using it in multiple places.

njhale (approved):

pkg/runner/runner.go
g-linville: Good call

g-linville: Fixed. Decided to just define a function.

thedadams (approved):

PR #252, opened 2024-04-16
feat: warn when expected API key is missing

kaihendry:
I'm not sure why you set it as "\<unset\>" here, can you please explain?


ibuildthecloud:
There are providers, mostly local providers (LocalAI, ollama, LMStudio, etc), that don't require a key but there are code paths in gptscript, or maybe the openai client we use, that require the key to be set. Thus the fake value.

We are going to extend the credential framework to handle these magic env vars, so the UX will be much more explicit and better. In the short term I'm fine with the log message.

ibuildthecloud (approved):

Issue #256, opened 2024-04-16
Credentials - When there are parallel tool calls made , there are texts relating to tool calls that gets displayed after the prompt (when waiting for user input) which is confusing. 


sangee2004:
gptscript version v0.0.0-dev-bc620f0e-dirty

**Steps to reproduce the problem:**
1. Run `examples/story-book/story-book.gpt` which makes  multiple parallel tool calls to `image-generation` tool.

2. Notice that  there are texts relating to tool calls that gets displayed after the prompt "?token" which is confusing. 

`gptscript --disable-cache examples/story-book/story-book.gpt --story "My trip to Japan" --pages 4`

```
13:04:29 sent     [story-illustrator(11)]
13:04:29 sent     [story-illustrator(9)]
13:04:29 sent     [story-illustrator(8)]
         content  [8] content | Waiting for model response...
13:04:29 sent     [story-illustrator(10)]
         content  [8] content | tool call image-generation -> {"prompt":"In Kyoto, Jamie wanders through ancient streets li ...
13:04:32 started  [credential: github.com/gptscript-ai/image-generation/credential]
13:04:32 sent     [credential: github.com/gptscript-ai/image-generation/credential]
Please enter your OpenAI API token.
? token          content  [8] content | tool call image-generation -> {"prompt":"In Kyoto, Jamie wanders through ancient streets li ...
         content  [8] content | tool call image-generation -> {"prompt":"In Kyoto, Jamie wanders through ancient streets li ...         content  [11] content | tool call image-generation -> {"prompt": "Jamie in Tokyo, surrounded by skyscrapers and arc ...
         content  [10] content | tool call image-generation -> {"prompt":"Jamie in the heart of the countryside, climbing mi ...
         content  [9] content | tool call image-generation -> {"prompt":"Jamie at the Sakura festival in a village, surroun ...
         content  [9] content | tool call image-generation -> {"prompt":"Jamie at the Sakura festival in a village, surroun ...
***************************************************
13:06:47 ended    [credential: github.com/gptscript-ai/image-generation/credential]
13:06:47 started  [story-illustrator(11)->image-generation(13)] [input={"prompt": "Jamie in Tokyo, surrounded by skyscrapers and arcade games. In a pointilism cartoon children's book style with no text in it", "quality": "hd"}]
```

**Expected Behavior:**
Prevent the text relating to tool calls to be displayed after the prompt when waiting for users input.

sangee2004:
[...]

PR #261, opened 2024-04-17
fix: credentials: block output while running credential tools

g-linville:
for #256

One of the goals of the credentials framework was to block output to the terminal while running `sys.prompt`, since it prompts the user directly in the terminal. I originally implemented this by storing the Monitor into the Context and then pulling it out of the Context and running its `Pause` function in the implementation of `sys.prompt`. However, this does not work, as credential tools pretty much always have to call `sys.prompt` with a new gptscript subprocess.

This new implementation instead stores the Pause function inside of the context and calls it within the Engine code prior to running a credential tool. This way, the Monitor is paused in the correct process.

### DIFF
--- a/pkg/engine/cmd.go
+++ b/pkg/engine/cmd.go
@@ -20,7 +20,7 @@ import (
 	"github.com/gptscript-ai/gptscript/pkg/version"
 )
 
-func (e *Engine) runCommand(ctx context.Context, tool types.Tool, input string) (cmdOut string, cmdErr error) {
+func (e *Engine) runCommand(ctx context.Context, tool types.Tool, input string, pauseF func() func()) (cmdOut string, cmdErr error) {
 	id := fmt.Sprint(atomic.AddInt64(&completionID, 1))
 
 	defer func() {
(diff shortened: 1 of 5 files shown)


g-linville:
Talked with Darren and I'm gonna refactor this a bit.

pkg/engine/cmd.go
```diff
@@ -64,6 +65,12 @@ func (e *Engine) runCommand(ctx context.Context, tool types.Tool, input string)
 	cmd.Stderr = io.MultiWriter(all, os.Stderr)
 	cmd.Stdout = io.MultiWriter(all, output)
 
+	if isCredential {
+		pause := context2.GetPauseFuncFromCtx(ctx)
+		unpause := pause()
+		defer unpause()
```
g-linville, line 71: This could technically be simplified down to `defer context2.GetPauseFuncFromCtx(ctx)()()` but I was afraid that would be too confusing, so I split it into three lines to make it clearer what is happening here.

thedadams: Part of me would love to see this as `defer context2.GetPauseFuncFromCtx(ctx)()()`, but I think the split is the right idea.

thedadams (approved):

pkg/engine/cmd.go
ibuildthecloud: And then you'd be a terrible person. What maniac would do a chained function call in a single line defer :)

njhale (approved):

PR #269, opened 2024-04-19
chore: make the event context a simplified version of the engine context

thedadams:
This change will allow the unmarshaling of an event because it no longer contains the engine context.


pkg/engine/engine.go
```diff
@@ -53,7 +53,16 @@ type CallResult struct {
...
 
+type commonContext struct {
+	ID           string         `json:"id"`
+	Tool         types.Tool     `json:"tool"`
+	InputContext []InputContext `json:"inputContext"`
+	// IsCredential indicates that the current call is for a credential tool
+	IsCredential bool `json:"isCredential"`
+}
+
 type Context struct {
```
g-linville, line 64: Is there any need to retain the `ID`, `Tool`, `InputContext`, and `IsCredential` fields on this main Context type, now that they are in the embedded struct commonContext?

g-linville (approved):

njhale (approved):

PR #273, opened 2024-04-22
chore: add Windows ARM64 build

thedadams:


.github/workflows/release.yaml
```diff
@@ -40,5 +40,6 @@ jobs:
             Invoke-WebRequest -Uri 'https://aka.ms/wingetcreate/latest' -OutFile 'wingetcreate.exe'      
       - name: Create WinGet Package Update Pull Request
         run: |
-            $url = "${{ github.server_url }}/${{ github.repository }}/releases/download/${{ github.ref_name }}/gptscript-${{ github.ref_name }}-windows-amd64.zip"
-            ./wingetcreate.exe update --submit --token "${{ secrets.WINGET_GH_TOKEN }}" --urls $url --version "${{ github.ref_name }}" gptscript-ai.gptscript
+            $amdurl = "${{ github.server_url }}/${{ github.repository }}/releases/download/${{ github.ref_name }}/gptscript-${{ github.ref_name }}-windows-amd64.zip|x86"
+            $armurl = "${{ github.server_url }}/${{ github.repository }}/releases/download/${{ github.ref_name }}/gptscript-${{ github.ref_name }}-windows-arm64.zip|x64"
+            ./wingetcreate.exe update --submit --token "${{ secrets.WINGET_GH_TOKEN }}" --urls $amdurl $armurl --version "${{ github.ref_name }}" gptscript-ai.gptscript
```
thedadams, line 45: This is taken from an example from the `wingetcreate` documentation: `wingetcreate update <PackageIdentifier> --urls "<InstallerUrl1>|x64" "<InstallerUrl2>|x86"`

drpebcak (approved):

.github/workflows/release.yaml
drpebcak: I approved this and then immediately wondered, are these actually the correct arch?

drpebcak: The usage doc seems to indicate that this will fail because you can't add a new installer with the update command? https://github.com/microsoft/winget-create/blob/main/doc/update.md#usage

thedadams: Yes, confirm you are correct about this. For now, I will change this PR to create the Windows ARM64 binary, but not publish it until we find it necessary to do so.

drpebcak (approved):

PR #275, opened 2024-04-23
chore: include gptscript-credential-helpers in packaging

g-linville:
This should be the last step for packaging the GPTScript credential helpers, which are small CLI programs (forked from the docker-credential-helpers repo) that gptscript uses to interact with OS keystores. The source for this is at github.com/gptscript-ai/gptscript-credential-helpers

I recently revamped and cleaned up the credential helper repo and created its first GitHub release, v0.1.0. I also combined the amd64 and arm64 darwin binaries into a single universal binary using github.com/g-linville/makefat, my fork of the tool github.com/randall77/makefat. (We can move this fork to gptscript-ai if we want. I basically just added CI and a goreleaser config for it, and set up reproducible builds.)

These changes add hooks to the goreleaser file to download the compiled binaries for gptscript-credential-osxkeychain and gptscript-credential-wincred.exe from the v0.1.0 release in the credential helpers repo. Then, the programs get copied into the correct release folder for their corresponding OS+arch.

The trouble with this is that I can't fully test to make sure that the changes are picked up and working in Homebrew and Winget until this is in. Since Winget is just uploading the archive that contains the binaries, I think it should work. And I added another command to the Homebrew formula that should cause it to install the osxkeychain helper alongside gptscript itself.

### DIFF
--- a/scripts/copy-cred-helper.sh
+++ b/scripts/copy-cred-helper.sh
@@ -1,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+OS="$1"
+ARCH="$2"
+
+case "$OS" in
+	"darwin")
+		case "$ARCH" in
+			"amd64")
+				cp binaries/gptscript-credential-osxkeychain releases/default_darwin_amd64_v1
+				;;
+			"arm64")
+				cp binaries/gptscript-credential-osxkeychain releases/default_darwin_arm64
+				;;
+		esac
+		;;
+	"windows")
+		case "$ARCH" in
+			"amd64")
+				cp binaries/gptscript-credential-wincred.exe releases/default_windows_amd64_v1
+				;;
+		esac
+		;;
+esac
--- a/scripts/copy-cred-helper.sh
+++ b/scripts/copy-cred-helper.sh
@@ -19,7 +19,10 @@ case "$OS" in
 	"windows")
 		case "$ARCH" in
 			"amd64")
-				cp binaries/gptscript-credential-wincred.exe releases/default_windows_amd64_v1
+				cp binaries/gptscript-credential-wincred-amd64.exe releases/default_windows_amd64_v1/gptscript-credential-wincred.exe
+				;;
+			"arm64")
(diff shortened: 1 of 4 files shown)


scripts/copy-cred-helper.sh
```diff
@@ -0,0 +1,26 @@
...
+				cp binaries/gptscript-credential-osxkeychain releases/default_darwin_amd64_v1
+				;;
+			"arm64")
+				cp binaries/gptscript-credential-osxkeychain releases/default_darwin_arm64
+				;;
+		esac
+		;;
+	"windows")
+		case "$ARCH" in
+			"amd64")
```
thedadams, line 21: We did add `arm64` binary for Windows. Should we handle that here?

g-linville: Oh, yes I will be sure to add that. Thanks

scripts/download-cred-helpers.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# This script downloads the gptscript-credential-helpers. (For use in CI.)
+
+GPTSCRIPT_CRED_HELPERS_VERSION="v0.1.0"
+BINARY_DIR="binaries"
+
+mkdir -p "$BINARY_DIR"
+cd "$BINARY_DIR"
+
+wget -O gptscript-credential-osxkeychain "https://github.com/gptscript-ai/gptscript-credential-helpers/releases/download/${GPTSCRIPT_CRED_HELPERS_VERSION}/gptscript-credential-osxkeychain"
```
StrongMonkey, line 13: Why wget :)

StrongMonkey (approved):

scripts/copy-cred-helper.sh
g-linville: Added

scripts/download-cred-helpers.sh
g-linville: I actually like wget a lot just for downloading files, lol. I can never remember the arguments I need to give to curl. 🤷‍♂️

thedadams (approved):

PR #289, opened 2024-04-26
bug: fix chat prompter missing format

rinor:


ibuildthecloud:
Ah, good catch, thanks!

PR #305, opened 2024-04-30
enhance: update credentials framework for OAuth support

g-linville:
This adds support for expiration and refresh tokens to the credential framework. All of these changes are backwards compatible with existing credentials, though new credentials will be stored in a slightly different format.

Notable changes:
- An expiration time for the credential will be saved if the cred tool provides that field. Once the credential is expired, the next time it is requested, the credential tool will be run again.
- A refresh token for the credential will be saved if the cred tool provides that field. This is useful in certain OAuth situations for requesting a new token without user interaction.
- `gptscript cred` output looks a little nicer and displays more information.

### DIFF
--- a/pkg/cli/credential.go
+++ b/pkg/cli/credential.go
@@ -93,7 +93,7 @@ func (c *Credential) Run(_ *cobra.Command, _ []string) error {
 		expires := "never"
 		if cred.ExpiresAt != nil {
 			if !cred.IsExpired() {
-				expires = cred.ExpiresAt.Sub(time.Now()).Truncate(time.Second).String()
+				expires = time.Until(*cred.ExpiresAt).Truncate(time.Second).String()
 			} else {
 				expires = "expired"
 			}
--- a/pkg/cli/credential.go
+++ b/pkg/cli/credential.go
@@ -6,6 +6,7 @@ import (
 	"sort"
 	"strings"
 	"text/tabwriter"
+	"time"
 
 	cmd2 "github.com/acorn-io/cmd"
 	"github.com/gptscript-ai/gptscript/pkg/cache"
@@ -46,6 +47,7 @@ func (c *Credential) Run(_ *cobra.Command, _ []string) error {
 	}
 	opts.Cache = cache.Complete(opts.Cache)
 
+	// Initialize the credential store and get all the credentials.
 	store, err := credentials.NewStore(cfg, ctx, opts.Cache.CacheDir)
 	if err != nil {
 		return fmt.Errorf("failed to get credentials store: %w", err)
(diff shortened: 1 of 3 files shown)


pkg/cli/credential.go
```diff
@@ -92,24 +83,49 @@ func (c *Credential) Run(_ *cobra.Command, _ []string) error {
...
+	}
+
+	for _, cred := range creds {
+		expires := "never"
+		if cred.ExpiresAt != nil {
+			if !cred.IsExpired() {
+				expires = time.Until(*cred.ExpiresAt).Truncate(time.Second).String()
+			} else {
+				expires = "expired"
 			}
```
njhale, line 99: style:

```suggestion
            expires = "expired"
			if !cred.IsExpired() {
				expires = time.Until(*cred.ExpiresAt).Truncate(time.Second).String()
			}
```

pkg/cli/credential.go
```diff
@@ -92,24 +83,49 @@ func (c *Credential) Run(_ *cobra.Command, _ []string) error {
...
-				sort.Strings(envVars)
-				_, _ = fmt.Fprintf(w, "%s\t%s\n", cred.ToolName, strings.Join(envVars, ", "))
+			_, _ = w.Write([]byte("CREDENTIAL\tEXPIRES IN\tENV\n"))
+		} else {
+			_, _ = w.Write([]byte("CREDENTIAL\tEXPIRES IN\n"))
+		}
+	}
+
+	for _, cred := range creds {
+		expires := "never"
```
njhale, line 93: nit: may want to use constants here

njhale (approved):
Left a few nits, but otherwise LGTM!

pkg/credentials/credential.go
```diff
@@ -4,43 +4,58 @@ import (
...
 		ServerAddress: toolNameWithCtx(c.ToolName, c.Context),
 	}, nil
 }
 
 func credentialFromDockerAuthConfig(authCfg types.AuthConfig) (Credential, error) {
-	var env map[string]string
-	if err := json.Unmarshal([]byte(authCfg.Password), &env); err != nil {
-		return Credential{}, err
+	var cred Credential
+	if err := json.Unmarshal([]byte(authCfg.Password), &cred); err != nil || len(cred.Env) == 0 {
```
StrongMonkey, line 52: Do we intentionally ignore error here?

pkg/cli/credential.go
```diff
@@ -65,25 +71,10 @@ func (c *Credential) Run(_ *cobra.Command, _ []string) error {
...
-
-			for _, cred := range creds {
-				envVars := make([]string, 0, len(cred.Env))
-				for envVar := range cred.Env {
-					envVars = append(envVars, envVar)
-				}
-				sort.Strings(envVars)
-				_, _ = fmt.Fprintf(w, "%s\t%s\t%s\n", cred.Context, cred.ToolName, strings.Join(envVars, ", "))
-			}
+			_, _ = w.Write([]byte("CONTEXT\tCREDENTIAL\tEXPIRES IN\tENV\n"))
```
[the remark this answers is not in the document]
g-linville: Yep. golangci-lint complains, and GoLand underlines it

pkg/credentials/credential.go
g-linville: We're not really ignoring the error. We used to store just a `map[string]string` in the credential, but now we store the entire JSON credential object. So if the initial JSON marshal fails, we try unmarshaling just the environment variables. Then if that fails, we return an error.

PR #309, opened 2024-05-01
fix: stop discarding first line shebangs

thedadams:
The code explicitly discards the first line of a gptscript file if it starts with a shebang. However, if the first line is blank, then the shebang is kept as expected.

This is inconsistent behavior that this change attempts to rectify.


drpebcak (approved):

drpebcak:
no idea why this is the way it is, but i support changing it!

thedadams:
Closing because someone has chimed in with the reason it is done this way. Will reopen if a desirable solution is reached.

Issue #311, opened 2024-05-03
chat resume failure


yglcode:
Reproduce step:
1. export GPTSCRIPT_CHAT_STATE=null; and run a chat.gpt script and save the returned initial state to GPTSCRIPT_CHAT_STATE.
2. rerun the chat.gpt script should be able to resume; however resume failed.

should the following line marshal resp.State in stead of resp?
data, err := json.Marshal(resp.State)
https://github.com/gptscript-ai/gptscript/blob/52d6974ab7e9d654ed475af9267ce5599dc1c0a5/pkg/cli/gptscript.go#L386

Thanks

thedadams:
Hi @yglcode! Thanks for your question.

The current implementation is operating as exampled. The `resp` will contain a few things, one of which is the chat state. In your case, `GPTSCRIPT_CHAT_STATE` would need to be the state, but the rest of the information in `resp` is valuable.

Here is a simple example script to show the intended use. Using the `--chat-state` command line argument is equivalent to setting the `GPTSCRIPT_CHAT_STATE` environment variable.

```bash
#!/bin/bash

RESP="$(gptscript --chat-state null chat.gpt Starter message)"

while true; do
    STATE=$(echo "$RESP" | jq .state)
    DONE=$(echo "$RESP" | jq .done)
    CONTENT=$(echo "$RESP" | jq .content)

    if [ -n "$CONTENT" ]; then
        echo "$CONTENT"
    fi

    if [ "$DONE" == "true" ]; then
        exit
    fi

    echo -n "> "
    read INPUT

    RESP="$(gptscript --chat-state "$STATE" chat.gpt "$INPUT")"
done
```

yglcode:
Hi @thedadams , thanks for the detailed explanation and help! i am confused by the cmdline doc. it says "... 
 return the state":
--chat-state string            The chat state to continue, or null to start a new chat and return the state ($GPTSCRIPT_CHAT_STATE)

PR #313, opened 2024-05-03
chore: prevent panics on invalid state input

rinor:
added just enough checks to prevent some panics on invalid chat state inputs encountered while investigating https://github.com/gptscript-ai/gptscript/issues/311


ibuildthecloud:
Did you hit a situation where you actually got a panic? I'm fine with this change, just curious if there's a bug elsewhere we need to fix.

ibuildthecloud (approved):

PR #337, opened 2024-05-09
fix: set encoding for python tools to utf-8

drpebcak:


thedadams (approved):

njhale (approved):

pkg/repos/runtimes/python/python.go
```diff
@@ -134,6 +134,9 @@ func (r *Runtime) Setup(ctx context.Context, dataRoot, toolSource string, env []
 	}
 
 	newEnv := runtimeEnv.AppendPath(env, venvBinPath)
+	if runtime.GOOS == "windows" && os.Getenv("PYTHONIOENCODING") != "" {
```
thedadams, line 137: Ah, crap. I didn't catch this before, but this should be `== ""`

thedadams (changes requested):

pkg/repos/runtimes/python/python.go
drpebcak: Oh duh. Pair programming made us worse haha.

thedadams (approved):

PR #340, opened 2024-05-10
bug: speed up the openapi loading

ibuildthecloud:


cloudnautique (approved):

pkg/loader/github/github.go
```diff
@@ -37,7 +40,14 @@ func getCommitLsRemote(ctx context.Context, account, repo, ref string) (string,
 	return git.LsRemote(ctx, url, ref)
 }
 
+// regexp to match a git commit id
+var commitRegexp = regexp.MustCompile("^[a-z0-9]{40}$")
```
rinor, line 44:
```suggestion
var commitRegexp = regexp.MustCompile("^[a-f0-9]{40}$")
```

ibuildthecloud: thanks, good catch

PR #341, opened 2024-05-10
fix: tool names should be used instead of filename

cloudnautique:
Uses the toolname before the from otherwise it was using the filename and appending numbers losing context.

### DIFF
--- a/pkg/types/toolname.go
+++ b/pkg/types/toolname.go
@@ -3,6 +3,7 @@ package types
 import (
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strings"
 
 	"github.com/gptscript-ai/gptscript/pkg/system"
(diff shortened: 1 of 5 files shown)


pkg/types/toolname.go
```diff
@@ -14,7 +14,8 @@ var (
 )
 
 func ToolNormalizer(tool string) string {
-	parts := strings.Split(tool, "/")
+	fromParts := strings.Split(tool, "from")
```
ibuildthecloud, line 17: There a method `SplitToolRef` for this

ibuildthecloud: It will make the code a bit more complicated, but it will be more correct.

pkg/types/toolname.go
```diff
@@ -14,7 +15,8 @@ var (
 )
 
 func ToolNormalizer(tool string) string {
-	parts := strings.Split(tool, "/")
+	_, subTool := SplitToolRef(tool)
```
rinor, line 18:
```suggestion
	toolName, subTool := SplitToolRef(tool)
	if subTool == "" {
		subTool = toolName
	}
```

ibuildthecloud (approved):

PR #344, opened 2024-05-14
fix: do not lowercase all items in tool(like) lists

cloudnautique:
When importing things from github or other case sensitive location the previous functionality would lowercase everything causing clones from VCS to fail.

This fix removes the calls to lowercase tools and contexts along with the other places they occur.

### DIFF
--- a/pkg/loader/github/github.go
+++ b/pkg/loader/github/github.go
@@ -41,7 +41,7 @@ func getCommitLsRemote(ctx context.Context, account, repo, ref string) (string, 
 }
 
 // regexp to match a git commit id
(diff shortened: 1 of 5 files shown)


pkg/loader/github/github.go
```diff
@@ -41,7 +41,7 @@ func getCommitLsRemote(ctx context.Context, account, repo, ref string) (string,
 }
 
 // regexp to match a git commit id
-var commitRegexp = regexp.MustCompile("^[a-f0-9]{40}$")
+var commitRegexp = regexp.MustCompile("^[a-z0-9]{40}$")
```
ibuildthecloud, line 44: i don't think you meant to make this change

PR #345, opened 2024-05-14
add: OpenAPI v2 tools support

iwilltry42:
Support OpenAPI v2 spec for tools by converting them up to v3 and re-using the existing v3 logic.

Is Yaml? -> Transform to JSON -> Is JSON? -> Unmarshal to OpenAPI v2 spec -> convert to OpenAPI v3 spec -> treat with existing v3 logic.

### DIFF
--- a/pkg/loader/loader.go
+++ b/pkg/loader/loader.go
@@ -148,37 +148,33 @@ func loadOpenAPI(prg *types.Program, data []byte) *openapi3.T {
 		prg.OpenAPICache = map[string]any{}
 	}
 
-	if ver, ok := isOpenAPI(data); ok {
-		switch ver {
-		case 2:
-			// Convert OpenAPI v2 to v3
-			jsondata := data
-			if !json.Valid(data) {
-				jsondata, err = kyaml.YAMLToJSON(data)
-				if err != nil {
-					return nil
-				}
-			}
-
-			doc := &openapi2.T{}
-			if err := doc.UnmarshalJSON(jsondata); err != nil {
-				return nil
-			}
-
-			openAPIDocument, err = openapi2conv.ToV3(doc)
+	switch isOpenAPI(data) {
+	case 2:
+		// Convert OpenAPI v2 to v3
+		jsondata := data
+		if !json.Valid(data) {
+			jsondata, err = kyaml.YAMLToJSON(data)
 			if err != nil {
 				return nil
 			}
-		case 3:
-			// Use OpenAPI v3 as is
-			openAPIDocument, err = openapi3.NewLoader().LoadFromData(data)
-			if err != nil {
-				return nil
-			}
-		default:
+		}
+
+		doc := &openapi2.T{}
+		if err := doc.UnmarshalJSON(jsondata); err != nil {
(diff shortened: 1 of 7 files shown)


g-linville (approved):

iwilltry42:
Rebased on top of https://github.com/gptscript-ai/gptscript/pull/340

iwilltry42:
Rebased once more

pkg/loader/loader.go
```diff
@@ -12,10 +12,14 @@ import (
 	"path"
 	"path/filepath"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 	"unicode/utf8"
 
+	"github.com/getkin/kin-openapi/openapi2"
+
```
thedadams, line 21: nit:
```suggestion
```

pkg/loader/loader.go
```diff
@@ -142,9 +148,38 @@ func loadOpenAPI(prg *types.Program, data []byte) *openapi3.T {
...
 	}
 
-	openAPIDocument, err = openapi3.NewLoader().LoadFromData(data)
-	if err != nil || openAPIDocument.Paths.Len() == 0 {
-		openAPIDocument = nil
+	if ver, ok := isOpenAPI(data); ok {
+		switch ver {
+		case 2:
+			// Convert OpenAPI v2 to v3
+			jsondata := data
```
thedadams, line 155: Not sure if this is intentional, but maybe this is more readable?
```suggestion
			var jsondata []byte
```

pkg/loader/loader.go
```diff
@@ -399,3 +434,42 @@ func input(ctx context.Context, cache *cache.Client, base *source, name string)
 
 	return nil, fmt.Errorf("can not load tools path=%s name=%s", base.Path, name)
 }
+
+// isOpenAPI checks if the data is an OpenAPI definition and returns the version if it is.
+func isOpenAPI(data []byte) (int, bool) {
```
thedadams, line 439: You can simplify the signature of this function. Returning `0` or `-1` to mean "not OpenAPI" seems valid. It would also make the above if-and-switch statement a little simpler.

pkg/loader/loader_test.go
```diff
@@ -5,6 +5,10 @@ import (
 	"encoding/json"
 	"testing"
 
+	"os"
+
+	"github.com/gptscript-ai/gptscript/pkg/types"
+
```
thedadams, line 11: nit: move `"os"` up with the other standard library packages and put `"github.com/gptscript-ai/gptscript/pkg/types"` with the other non-standard library packages.

iwilltry42: no idea, why those were not sorted automatically...

pkg/loader/loader.go
iwilltry42: Makes sense, now that it's only used once

pkg/loader/loader.go
thedadams: This line should still be removed.

thedadams (approved):

PR #364, opened 2024-05-16
chore: change sys.download to use the workspace by default

g-linville:


pkg/builtin/builtin.go
```diff
@@ -162,7 +162,7 @@ var tools = map[string]types.Tool{
 			Description: "Downloads a URL, saving the contents to disk at a given location",
 			Arguments: types.ObjectSchema(
 				"url", "The URL to download, either http or https.",
-				"location", "(optional) The on disk location to store the file. If no location is specified a temp location will be used. If the target file already exists it will fail unless override is set to true.",
+				"location", "(optional) The on disk location to store the file. If no location is specified, the workspace will be used. If the target file already exists, it will fail unless override is set to true.",
```
ibuildthecloud, line 165: I wouldn't change the description. Temp location is still technically correct and doesn't assume the LLM is aware of what "the workspace" means

PR #383, opened 2024-05-23
Add link to the definition of shebangs to readme.

arthurwolf:


cjellick:
Thanks for the suggestion. I've decided to not merge. While I realize #! might be a bit esoteric, so is the explanation in that article. I don't think adding the link clarifies anything.

PR #387, opened 2024-05-24
feat: add a new SDK server

thedadams:
This server will be used in the SDKs for running gptscripts.

### DIFF
--- a/pkg/engine/daemon.go
+++ b/pkg/engine/daemon.go
@@ -14,6 +14,8 @@ import (
 	"github.com/gptscript-ai/gptscript/pkg/types"
 )
 
+var ports Ports
+
 type Ports struct {
 	daemonPorts map[string]int64
 	daemonLock  sync.Mutex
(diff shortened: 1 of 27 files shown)


pkg/sdkserver/routes.go
```diff
@@ -0,0 +1,231 @@
...
+	logger := gcontext.GetLogger(r.Context())
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		writeError(logger, w, http.StatusInternalServerError, fmt.Errorf("failed to read request body: %w", err))
+		return
+	}
+
+	if body[len(body)-1] != '}' {
+		body = append(body, '"', '}')
+	}
```
g-linville, line 141: Is this in case a model making calls through the SDK fails to finish a JSON string?

pkg/sdkserver/server.go
```diff
@@ -0,0 +1,93 @@
...
+		return err
+	}
+
+	s := &server{
+		client: g,
+		events: events,
+	}
+	defer s.Close()
+
+	s.addRoutes(http.DefaultServeMux)
```
g-linville, line 59: Where does this `s` server end up getting used?

pkg/engine/daemon.go
```diff
@@ -93,26 +99,31 @@ func getPath(instructions string) (string, string) {
...
+	ports.daemonLock.Lock()
+	defer ports.daemonLock.Unlock()
 
 	instructions := strings.TrimPrefix(tool.Instructions, types.DaemonPrefix)
 	instructions, path := getPath(instructions)
 	tool.Instructions = types.CommandPrefix + instructions
 
-	port, ok := e.Ports.daemonPorts[tool.ID]
+	port, ok := ports.daemonPorts[tool.ID]
 	url := fmt.Sprintf("http://127.0.0.1:%d%s", port, path)
```
iwilltry42, line 111: Not sure if there would be any use for this (and it's not part of your PR anyway), but can this be made configurable to listen on other interfaces as well?

pkg/sdkserver/routes.go
thedadams: This was supposed to be removed.

pkg/sdkserver/server.go
thedadams: In `addRoutes`, it sets up the routes for the various commands, which is where the client and events get used.

pkg/engine/daemon.go
thedadams: We can definitely discuss that!

pkg/sdkserver/routes.go
```diff
@@ -0,0 +1,227 @@
...
+// parse will parse the file and return the corresponding Document.
+func (s *server) parse(w http.ResponseWriter, r *http.Request) {
+	logger := gcontext.GetLogger(r.Context())
+	reqObject := new(parseRequest)
+	if err := json.NewDecoder(r.Body).Decode(reqObject); err != nil {
+		writeError(logger, w, http.StatusBadRequest, fmt.Errorf("invalid request body: %w", err))
+		return
+	}
+
+	logger.Debugf("parsing file: file=%s, content=%s", reqObject.File, reqObject.Content)
```
iwilltry42, line 188: to log with fields or not to log with fields? :grin:

pkg/sdkserver/run.go
```diff
@@ -0,0 +1,172 @@
...
+		if f, ok := w.(http.Flusher); ok {
+			f.Flush()
+		}
+	}
+
+	logger.Debugf("wrote DONE event")
+}
+
+// streamEvents will stream the events of the tool to the response as server sent events.
+// This looks for and tries to handle confirm events as well. However, that currently is not implemented in the SDK.
```
iwilltry42, line 85: What's the confirm event below?

pkg/sdkserver/types.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+type output struct {
+	Content  string                 `json:"content"`
+	SubCalls map[string]engine.Call `json:"subCalls"`
+}
+
+type runState string
+
+const (
```
iwilltry42, line 181: Maybe the constants should go to the top of the file?

iwilltry42 (approved):
LGTM - no blocking comments, just some questions for my understanding :)

pkg/sdkserver/run.go
thedadams: This changed from the initial implementation. I removed the second line of this comment.

pkg/sdkserver/types.go
thedadams: Done.

PR #396, opened 2024-05-29
enhance: ask user for OpenAI key and store it in the cred store

g-linville:


thedadams (approved):

pkg/gptscript/gptscript.go
```diff
@@ -65,7 +66,12 @@ func New(opts *Options) (*GPTScript, error) {
 		return nil, err
 	}
 
-	oAIClient, err := openai.NewClient(opts.OpenAI, openai.Options{
+	cliCfg, err := config.ReadCLIConfig(opts.OpenAI.ConfigFile)
+	if err != nil {
+		return nil, err
+	}
+
+	oAIClient, err := openai.NewClient(cliCfg, opts.CredentialContext, opts.OpenAI, openai.Options{
```
njhale, line 74: nit:

```suggestion
	oaiClient, err := openai.NewClient(cliCfg, opts.CredentialContext, opts.OpenAI, openai.Options{
```

njhale (approved):

g-linville:
I made some pretty substantial changes to this PR:

- We now use SysPrompt to ask the user (which means I had to move it to its own package in order to avoid an import cycle)
- We only prompt the user for the token when it is actually needed
  - This involved creating a custom error type and checking for it in a couple places, which is a bit gross but is the cleanest solution I could come up with for this

pkg/prompt/prompt.go
```diff
@@ -0,0 +1,88 @@
...
+	data, err = io.ReadAll(resp.Body)
+	return string(data), err
+}
+
+func SysPrompt(ctx context.Context, envs []string, input string) (_ string, err error) {
+	var params struct {
+		Message   string `json:"message,omitempty"`
+		Fields    string `json:"fields,omitempty"`
+		Sensitive string `json:"sensitive,omitempty"`
+	}
```
StrongMonkey, line 48:
```suggestion
	var params types.Prompt
```

StrongMonkey: Looks like this can reuse the existing types.Prompt

StrongMonkey (approved):

pkg/prompt/prompt.go
StrongMonkey: Oh I see, then nvm

cjellick:
We will need to make sure this works with @tylerslaton's UI, once he has prompting support

pkg/credentials/credential.go
```diff
@@ -8,9 +8,17 @@ import (
 	"github.com/docker/cli/cli/config/types"
 )
 
+type CredentialType string
+
+const (
+	CredentialTypeTool          CredentialType = "tool"
+	CredentialTypeModelProvider CredentialType = "modelprovider"
```
StrongMonkey, line 15: Nit: `modelProvider` might be better?

pkg/credentials/helper.go
```diff
@@ -55,7 +57,32 @@ func (h *HelperStore) GetAll() (map[string]types.AuthConfig, error) {
 		return nil, err
 	}
 
-	for serverAddress := range serverAddresses {
+	newCredAddresses := make(map[string]string, len(serverAddresses))
+	for serverAddress, val := range serverAddresses {
```
StrongMonkey, line 61: Interesting... that's weird behavior from mac library

pkg/credentials/credential.go
```diff
@@ -33,14 +41,28 @@ func credentialFromDockerAuthConfig(authCfg types.AuthConfig) (Credential, error
...
+	// We used to hardcode the username as "gptscript" before CredentialType was introduced, so
+	// check for that here.
+	credType := authCfg.Username
+	if credType == "gptscript" {
+		credType = string(CredentialTypeTool)
+	}
+
+	// If it's a tool credential or sys.openai, remove the http[s] prefix.
+	address := authCfg.ServerAddress
+	if credType == string(CredentialTypeTool) || strings.HasPrefix(address, "https://sys.openai///") {
```
StrongMonkey, line 53: Is `///` a typo or is it desired to be this way?

StrongMonkey (approved):

pkg/credentials/helper.go
g-linville: Yep it definitely took me a bit to figure that one out lol

pkg/credentials/credential.go
g-linville: fixed

pkg/credentials/credential.go
g-linville: It was intentional. I use it as the separator between the credential tool name/alias and the credential context, both of which get stored in the "server address" of the credential in the keychain. I created a const for this to add more clarity.

pkg/openai/client.go
```diff
@@ -121,8 +162,9 @@ func (c *Client) ListModels(ctx context.Context, providers ...string) (result []
 		return nil, nil
 	}
 
+	// If auth is invalid, we just want to return nothing.
```
thedadams, line 165: nit: could you amend the comment to explain why we want to do nothing?

thedadams (approved):

pkg/openai/client.go
g-linville: Done

thedadams (approved):

PR #418, opened 2024-06-03
Remove gen-docs make target

cjellick:
The target is broken - it is calling a gptscript that was deleted awhile
ago. We probably do want a new version of gen-docs, but this isn't it.
So, I'm deleting it for now.

Signed-off-by: Craig Jellick <craig@acorn.io>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,5 @@ validate: tidy lint
 ci: build
 	./bin/gptscript ./scripts/ci.gpt
 
-gen-docs: build
-	./bin/gptscript ./scripts/gen-docs.gpt
-
 serve-docs:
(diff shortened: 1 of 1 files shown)


drpebcak (approved):

ibuildthecloud:
lol, I deleted that a while ago. So do we just not do docs?

cjellick:
> lol, I deleted that a while ago. So do we just not do docs?

well, we def don't call gen-docs. I think ultimately we should re-introduce what we did with acorn and programatically generate docs for the cli flags and fail PRs if they get out of sync.

Actually @drpebcak - could you add this to your stack of tasks?

PR #428, opened 2024-06-04
fix: add ability to set required extensions for certain commands

thedadams:
Some commands, like powershell, require the files it runs to end with a particular extension. This change sets up this requirement with the ability to add more later.

### DIFF
--- a/pkg/engine/cmd.go
+++ b/pkg/engine/cmd.go
@@ -20,6 +20,11 @@ import (
 	"github.com/gptscript-ai/gptscript/pkg/version"
 )
 
+var requiredFileExtensions = map[string]string{
+	"powershell.exe": "*.ps1",
+	"powershell":     "*.ps1",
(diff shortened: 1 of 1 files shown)


pkg/engine/cmd.go
```diff
@@ -20,6 +20,11 @@ import (
 	"github.com/gptscript-ai/gptscript/pkg/version"
 )
 
+var requiredFileExtensions = map[string]string{
+	"powershell.exe": "*.ps1",
+	"powershell":     "*.ps1",
```
g-linville, line 25: Why are there asterisks on these? I'm a bit confused

thedadams: The `*` allows Go to add randomness in the temporary file name.

g-linville (approved):

njhale (approved):

PR #443, opened 2024-06-05
fix: add JSON tags to Repo type

thedadams:


thedadams:
Not going to do this.

PR #444, opened 2024-06-05
fix: pause confirm in the SDK server

thedadams:


pkg/sdkserver/confirm.go
```diff
@@ -14,6 +14,8 @@ import (
 )
 
 func (s *server) authorize(ctx engine.Context, input string) (runner.AuthorizerResponse, error) {
+	defer gcontext.GetPauseFuncFromCtx(ctx.Ctx)()()
```
thedadams, line 17: The idea is that we do similar things as this one: https://github.com/gptscript-ai/gptscript/blob/0a758945566a8281bdd024820bcb0c90fa46b577/pkg/auth/auth.go#L16

tylerslaton (approved):

PR #457, opened 2024-06-07
Add multi-model smoke tests

njhale:
Add a multi-model smoke test workflow to detect breaking changes to core GPTScript features.

The smoke test runner uses `gpt-4o` to perform a fuzzy equality check between existing golden files and event stream files generated at runtime.

To add a new test case, just create a new directory and the GPTScript you want to test; e.g. `pkg/tests/smoke/testdata/<test-case>/<script>.gpt`, set your `GPTSCRIPT_DEFAULT_MODEL` (and the respective auth environment variables), and run the `smoke` make target. This will generate the initial, model-specific, golden file in the `pkg/tests/smoke/testdata/<test-case>` directory. Successive runs will then reference this file for comparison.

In order for the workflow to run for PRs from external contributors -- that aren't members of the `gptscript-ai` org -- a member must first add the `run-smoke` label to the PR. This gate gives members a chance to review the PR first and ensure that the PR doesn't contain code that would compromise the org's GitHub secrets.


.github/workflows/smoke.yaml
```diff
@@ -0,0 +1,164 @@
...
+
+jobs:
+  check-label:
+    runs-on: ubuntu-22.04
+    outputs:
+      run_smoke_tests: ${{ steps.check.outputs.run_smoke_tests }}
+    steps:
+      - name: Check if PR author is a member of the organization or has the run-smoke label
+        id: check
+        run: |
```
g-linville, line 25: Is there a better way to gate this than running a script? Would the script cancel with an error if one of the curl requests gets something like a 500 back?

StrongMonkey (approved):

Makefile
```diff
@@ -14,6 +14,9 @@ tidy:
 test:
 	go test -v ./...
 
+smoke:
+	go test -v -tags='smoke' ./pkg/tests/smoke/...
```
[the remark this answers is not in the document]
njhale: I used a tag mainly to prevent the smoke tests from running on `go test -v ./...` (used by `make test`).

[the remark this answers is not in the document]
StrongMonkey: Oh, make sense 👍

.github/workflows/smoke.yaml
```diff
@@ -0,0 +1,164 @@
...
+          fetch-depth: 1
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.event.pull_request.head.ref }}
+      - uses: actions/setup-go@v5
+        with:
+          cache: false
+          go-version: "1.21"
+      - env:
+          OPENAI_API_KEY: ${{ secrets.SMOKE_OPENAI_API_KEY }}
+          GPTSCRIPT_SMOKE_DEFAULT_MODEL: claude-3-opus-20240229 from github.com/gptscript-ai/claude3-anthropic-provider@tool-beta
```
g-linville, line 129: Is this the correct env name or should it be `GPTSCRIPT_DEFAULT_MODEL`?

drpebcak (approved):

.github/workflows/smoke.yaml
njhale: You're correct. Thanks for catching that.

.github/workflows/smoke.yaml
njhale: This has been updated a bit since your initial comment. Any non `204` response code from the curl command will cause smoke tests to be skipped, but will not cause the check-labels job itself to fail. I can tinker with this some more so that the check-labels job does fail for that case if you think it's worthwhile.

g-linville: I don't think there's a need to do that. I just wanted to be as sure as possible that the tests themselves will get skipped. :+1:

PR #461, opened 2024-06-07
feat: add support for remote file parsing

tylerslaton:
## Summary

This PR adds support for remote file parsing. This will mainly be a nice feature for the SDKs as consumers of them can now parse file remotely prior to executing them. Namely, the UI will use this functionality.

## How
This is done by exporting `loader.ContentFromURL` as a new public function. The `input` package then uses `loader.ContentFromURL` in a new `input.FromURL` function. This allows all of the same loading logic to apply that GPTScript runs do while also leveraging the same cache that they do. Finally, the `parse` subcommand calls `input.FromURL` if a file load fails.

### DIFF
--- a/pkg/cli/parse.go
+++ b/pkg/cli/parse.go
@@ -2,10 +2,12 @@ package cli
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"strings"
 
 	"github.com/gptscript-ai/gptscript/pkg/input"
+	"github.com/gptscript-ai/gptscript/pkg/loader"
 	"github.com/gptscript-ai/gptscript/pkg/parser"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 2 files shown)


pkg/input/input.go
```diff
@@ -39,3 +41,7 @@ func FromFile(file string) (string, error) {
 
 	return "", nil
 }
+
+func FromURL(url string) (string, error) {
```
thedadams, line 45: It seems that this function isn't necessary and callers can just call `loader.ContentFromURL`.

pkg/cli/parse.go
```diff
@@ -26,12 +26,23 @@ func locationName(l string) string {
...
+		err     error
+	)
+
+	// Attempt to read the file first, if that fails, try to load the URL. Finally,
+	// return an error if both fail.
+	content, err = input.FromFile(args[0])
 	if err != nil {
-		return err
+		log.Debugf("failed to read file %s (due to %v) attempting to load the URL...", args[0], err)
+		content, err = input.FromURL(args[0])
```
thedadams, line 39: It's possible to get `"", nil` here. This indicates that `args[0]` isn't a URL or GitHub repo.

I think it makes sense to return the original error in this case. What do you think?

pkg/input/input.go
tylerslaton: I mainly wanted them all of these to exist in the same package. Thoughts on that?

pkg/cli/parse.go
tylerslaton: I agree and will add that.

pkg/input/input.go
thedadams: Not something I think is valuable, but feel free to keep it.

pkg/cli/parse.go
```diff
@@ -26,12 +28,27 @@ func locationName(l string) string {
...
+	// return an error if both fail.
+	content, err = input.FromFile(args[0])
 	if err != nil {
-		return err
+		log.Debugf("failed to read file %s (due to %v) attempting to load the URL...", args[0], err)
+		content, err = loader.ContentFromURL(args[0])
+		if err != nil {
+			return err
+		}
+		// If the content is empty and there was no error, this is not a remote file. Return the file error.
```
thedadams, line 45: nit: comment is out of date.

thedadams (approved):

cjellick (approved):

PR #468, opened 2024-06-10
enhance: automatically set up credential helpers

g-linville:
This PR adds functionality to GPTScript to automatically clone and build the GPTScript credential helpers. I also expanded the information in the docs to explain each of the available helpers. I have tested this on Windows, macOS, and Linux, but I would like for at least one of the reviewers to test it on his system as well.

### DIFF
--- a/pkg/credentials/noop.go
+++ b/pkg/credentials/noop.go
@@ -1,0 +1,19 @@
+package credentials
+
+type NoopStore struct{}
+
+func (s NoopStore) Get(_ string) (*Credential, bool, error) {
+	return nil, false, nil
+}
+
+func (s NoopStore) Add(_ Credential) error {
+	return nil
+}
+
+func (s NoopStore) Remove(_ string) error {
+	return nil
+}
+
+func (s NoopStore) List() ([]Credential, error) {
+	return nil, nil
+}
--- a/pkg/repos/get.go
+++ b/pkg/repos/get.go
@@ -62,8 +62,10 @@ func New(cacheDir string, runtimes ...Runtime) *Manager {
 }
 
 func (m *Manager) SetUpCredentialHelpers(ctx context.Context, cliCfg *config.CLIConfig, env []string) error {
-	helperName := cliCfg.CredentialsStore
-	suffix := ""
+	var (
+		helperName = cliCfg.CredentialsStore
+		suffix     string
+	)
 	if helperName == "wincred" {
 		suffix = ".exe"
 	}
--- a/pkg/repos/get.go
+++ b/pkg/repos/get.go
@@ -8,11 +8,18 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/BurntSushi/locker"
+	"github.com/gptscript-ai/gptscript/pkg/config"
+	"github.com/gptscript-ai/gptscript/pkg/credentials"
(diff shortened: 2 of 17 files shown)


pkg/credentials/noop.go
```diff
@@ -0,0 +1,19 @@
+package credentials
+
+type NoopStore struct{}
+
+func (s NoopStore) Get(_ string) (*Credential, bool, error) {
```
thedadams, line 5: nit, here and below:
```suggestion
func (NoopStore) Get(string) (*Credential, bool, error) {
```

pkg/repos/get.go
```diff
@@ -36,20 +43,81 @@ func (n noopRuntime) Setup(_ context.Context, _, _ string, _ []string) ([]string
...
-		runtimes:   runtimes,
+		storageDir:     root,
+		gitDir:         filepath.Join(root, "git"),
+		runtimeDir:     filepath.Join(root, "runtimes"),
+		credHelperDirs: credentials.GetCredentialHelperDirs(cacheDir),
+		runtimes:       runtimes,
+	}
+}
+
+func (m *Manager) SetUpCredentialHelpers(ctx context.Context, cliCfg *config.CLIConfig, env []string) error {
```
thedadams, line 64: Curious why this is necessary and we can't use the built-in Go runtime support.

thedadams (commented):
The building of the credential helpers here is different than I expected.

pkg/repos/get.go
g-linville: The credential helpers aren't packaged as a tool so I figured it made more sense to handle the building of them separately (especially since it's one repo, but with several binaries that can be built from it). This does ultimately use our Go runtime in order to build it.

Is there a refactoring you think I could do to simplify and remove some of this credential helper-specific build stuff? I'm open to changing it if you think it should be different.

thedadams: Can it be packaged as a tool? Then this code would just be running the tool, which would build it.

g-linville: Well we don't want to actually run it here, only build it. But I'll see what I can do

g-linville: Having tried some things and thought more about this, I do think it is best to keep it as it is in this PR. While it may decrease the amount of code if I were to package these as tools and find ways to build and then find the built binaries when the credential store code needs to use them, I think the resulting code would be less clear and harder to maintain. Handling these separately is best, imo.

thedadams: OK. Thanks for looking into it.

thedadams (approved):

pkg/repos/get.go
```diff
@@ -36,20 +43,81 @@ func (n noopRuntime) Setup(_ context.Context, _, _ string, _ []string) ([]string
...
+	if !needsBuild {
+		// Check for the existence of the gptscript-credential-osxkeychain binary.
+		// If it's there, we have no need to build it and can just return.
+		if _, err := os.Stat(filepath.Join(m.credHelperDirs.BinDir, "gptscript-credential-"+helperName+suffix)); err == nil {
+			return nil
+		}
+	}
+
+	// Find the Go runtime and use it to build the credential helper.
+	for _, runtime := range m.runtimes {
```
StrongMonkey, line 113: Does gptscript initialize the runtime repo when it starts?

StrongMonkey (approved):

PR #480, opened 2024-06-11
feat: add sdkserver support for remote file parsing

tylerslaton:
Add the ability to parse remote scripts in the sdkserver.

To test this, you can do the following after pulling the branch down locally:

```sh
go run main.go --debug --listen-address 127.0.0.1:9090 sdkserver
```

and in a separate tab:

```sh
curl http://localhost:9090/parse -d '{"file":"github.com/gptscript-ai/cli-demo"}'
```

### DIFF
--- a/pkg/input/input.go
+++ b/pkg/input/input.go
@@ -5,6 +5,8 @@ import (
 	"io"
 	"os"
 	"strings"
+
+	"github.com/gptscript-ai/gptscript/pkg/loader"
 )
 
 func FromArgs(args []string) string {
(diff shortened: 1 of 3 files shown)


pkg/input/input.go
```diff
@@ -39,3 +41,28 @@ func FromFile(file string) (string, error) {
 
 	return "", nil
 }
+
+// FromString takes a string that can be a file path or a URL to a file and returns the content of that file.
+func FromString(s string) (string, error) {
```
thedadams, line 46: nit:
```suggestion
func FromLocation(s string) (string, error) {
```

pkg/input/input.go
```diff
@@ -39,3 +41,28 @@ func FromFile(file string) (string, error) {
...
+	var (
+		content string
+		err     error
+	)
+
+	// Attempt to read the file first, if that fails, try to load the URL. Finally,
+	// return an error if both fail.
+	content, err = FromFile(s)
```
thedadams, line 54: nit:
```suggestion
	// Attempt to read the file first, if that fails, try to load the URL. Finally,
	// return an error if both fail.
	content, err := FromFile(s)
```

thedadams (approved):

PR #488, opened 2024-06-12
enhance: automatically use the generic credential tool for OpenAPI

g-linville:
This is a relatively simple change that just adds a `Credential:` line or two to each tool generated for an OpenAPI definition, if there is some supported security scheme for the particular operation (i.e. API key, HTTP bearer token, HTTP basic auth).

This works seamlessly with existing OpenAPI tool examples such as the DigitalOcean agent.

### DIFF
--- a/pkg/types/tool.go
+++ b/pkg/types/tool.go
@@ -411,7 +411,9 @@ func (t ToolDef) String() string {
 		_, _ = fmt.Fprintf(buf, "Internal Prompt: %v\n", *t.Parameters.InternalPrompt)
 	}
 	if len(t.Parameters.Credentials) > 0 {
-		_, _ = fmt.Fprintf(buf, "Credentials: %s\n", strings.Join(t.Parameters.Credentials, ", "))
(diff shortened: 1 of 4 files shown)


pkg/types/tool.go
```diff
@@ -411,7 +411,9 @@ func (t ToolDef) String() string {
 		_, _ = fmt.Fprintf(buf, "Internal Prompt: %v\n", *t.Parameters.InternalPrompt)
 	}
 	if len(t.Parameters.Credentials) > 0 {
-		_, _ = fmt.Fprintf(buf, "Credentials: %s\n", strings.Join(t.Parameters.Credentials, ", "))
+		for _, cred := range t.Parameters.Credentials {
+			_, _ = fmt.Fprintf(buf, "Credential: %s\n", cred)
+		}
```
g-linville, line 416: We were previously printing the credential tools all on one line, which is no longer supported. This change makes it print each on its own line, which is supported.

StrongMonkey (approved):

thedadams (approved):

PR #491, opened 2024-06-13
enhance: add credential show command

g-linville:
This adds a new command to show the environment variables that are stored in a credential.


pkg/cli/credential.go
```diff
@@ -68,7 +69,7 @@ func (c *Credential) Run(_ *cobra.Command, _ []string) error {
 		defer w.Flush()
 
 		if c.ShowEnvVars {
-			_, _ = w.Write([]byte("CONTEXT\tTOOL\tENVIRONMENT VARIABLES\n"))
```
g-linville, line 71: We previously referred to a credential name simply by the tool that created it, which is still the default, but now that aliases are supported as well, I changed the output format here and in other places.

StrongMonkey (approved):

thedadams (approved):

PR #493, opened 2024-06-13
fix: support passing along URLs to the UI

tylerslaton:
This is necessary in combination with a change to the UI tool that will allow URLs to be passed through to the `run` routes `file` query param.

### DIFF
--- a/pkg/cli/gptscript.go
+++ b/pkg/cli/gptscript.go
@@ -338,17 +338,31 @@ func (r *GPTScript) Run(cmd *cobra.Command, args []string) (retErr error) {
 				return fmt.Errorf("chat UI only supports files, cannot read from stdin")
 			}
 
-			absPathToScript, err := filepath.Abs(args[1])
-			if err != nil {
-				return fmt.Errorf("cannot determine absolute path to script %s: %v", args[1], err)
+			file := args[1]
+
+			// If the file is external, then set the SCRIPTS_PATH to the current working directory. Otherwise,
+			// set it to the directory of the script and set the file to the base.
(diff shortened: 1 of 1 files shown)


pkg/cli/gptscript.go
```diff
@@ -338,17 +338,32 @@ func (r *GPTScript) Run(cmd *cobra.Command, args []string) (retErr error) {
...
 
-			absPathToScript, err := filepath.Abs(args[1])
-			if err != nil {
-				return fmt.Errorf("cannot determine absolute path to script %s: %v", args[1], err)
+			file := args[1]
+
+			// If the file is external, then set the SCRIPTS_PATH to the current working directory. Otherwise,
+			// set it to the directory of the script and set the file to the base.
+			external := strings.HasPrefix(file, "http://") || strings.HasPrefix(file, "https://") || strings.HasPrefix(file, "github.com")
+			if !external {
```
thedadams, line 346: nit, this is just my style. I don't like haven't variables that are used only once.
```suggestion
			external := strings.HasPrefix(file, "http://") || strings.HasPrefix(file, "https://") || strings.HasPrefix(file, "github.com")
			if !strings.HasPrefix(file, "http://") && !strings.HasPrefix(file, "https://") && !strings.HasPrefix(file, "github.com") {
```

pkg/cli/gptscript.go
```diff
@@ -338,17 +338,32 @@ func (r *GPTScript) Run(cmd *cobra.Command, args []string) (retErr error) {
...
+				absPathToScript, err := filepath.Abs(args[1])
```
thedadams, line 347: Since we have `file` above, can we use it here? I imagine someone changing the `file` variable and being confused with that change not being reflected here. Same as below on line 352.

thedadams (approved):

g-linville (approved):

Issue #483, opened 2024-06-11
Do not print credentials in console


sangee2004:
gptscript version v0.8.1+da33973d

**Steps to reproduce the problem:**
1. Have OpenAI API token already set (`sys.openai` credential created) by executing any gptscript.
2. Execute the following script that uses  `gpt4-v-vision` and `dalle-image-generation` tools.
```
Tools:  github.com/gptscript-ai/gpt4-v-vision, github.com/gptscript-ai/dalle-image-generation, sys.download

Generate an image of cow and download it in cow.png
Describe the image in file://./cow.png like a kindergarten teacher would.
```
3. Notice that when user is prompted for entering the credentials for `gpt4-v-vision` and `dalle-image-generation` tools , the credentials are printed in clear text in console
 
```
12:33:35 started  [vision(6)] [input={"images":"file://./cow.png","prompt":"Describe the image like a kindergarten teacher would."}]
12:33:35 started  [credential: github.com/gptscript-ai/gpt4-v-vision/credential]
12:33:35 sent     [credential: github.com/gptscript-ai/gpt4-v-vision/credential]
Please enter your OpenAI API token.
? token ***************************************************
12:34:06 ended    [credential: github.com/gptscript-ai/gpt4-v-vision/credential] [output={\"env\": {\"OPENAI_API_KEY\": \"sk-*******************\"}}] ==> This is in clear text
12:34:06 sent     [vision(6)]
12:34:19 ended    [vision(6)] [output={\n    \"id\": \"chatcmpl-9Z1Xv5M9by6w4lA34VOUnPj5zbepX\",\n    \"object\": \"chat.completion\",\n    \"created\"...]
```

sangee2004:
Tokens are also logged in clear text in "Stack Trace View" option when executing scripts using `Chat Builder` using "--ui" option.

PR #500, opened 2024-06-14
fix: redact output of credential tools

g-linville:
for #483

~~This creates a little wrapper type around the monitor, to be used only for credential calls, so that we do not print the plaintext output to the console.~~

Update: this instead excludes the output of credential tools from the CallFinish Event entirely, so the Monitor never has a chance to see it.

### DIFF
--- a/pkg/runner/monitor.go
+++ b/pkg/runner/monitor.go
@@ -33,17 +33,17 @@ type credWrapper struct {
 	m Monitor
 }
 
-func (n credWrapper) Event(e Event) {
+func (c credWrapper) Event(e Event) {
 	if e.Type == EventTypeCallFinish {
 		e.Content = "credential tool output redacted"
 	}
-	n.m.Event(e)
+	c.m.Event(e)
 }
 
-func (n credWrapper) Stop(s string, err error) {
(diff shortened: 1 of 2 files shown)


ibuildthecloud:
There might be a different way by just looking at the tool category in the monitor when it's printing.  This is effectively what the TUI does, we don't print output for tool category "context", "provider" or "credential"

ibuildthecloud:
Oh looking at this a big more, I think we probably need to drop the output in the place where we write the event.  The event should not have the clear text of a password in it.  Regardless if it's printed.  But I'm still lost on how the event even has the clear text...  Soo many layers. What madness have we created.

g-linville:
> Oh looking at this a big more, I think we probably need to drop the output in the place where we write the event. The event should not have the clear text of a password in it. Regardless if it's printed. But I'm still lost on how the event even has the clear text... Soo many layers. What madness have we created.

@ibuildthecloud I can figure this out. I will probably have to plumb some gross thing through all the layers to get it to work, but as you said, that's the nature of the madness.

ibuildthecloud (approved):

thedadams (approved):

PR #536, opened 2024-06-24
fix: misc improvements related to creds and prompting

g-linville:


pkg/prompt/prompt.go
```diff
@@ -75,6 +75,13 @@ func SysPrompt(ctx context.Context, envs []string, input string, _ chan<- string
 func sysPrompt(ctx context.Context, req types.Prompt) (_ string, err error) {
 	defer context2.GetPauseFuncFromCtx(ctx)()()
 
+	if req.Message != "" && len(req.Fields) == 1 && strings.TrimSpace(req.Fields[0]) == "" {
+		_, _ = fmt.Fprintln(os.Stderr, req.Message)
+		_, _ = fmt.Fprintln(os.Stderr, "Press enter to continue...")
+		_, _ = fmt.Fscanln(os.Stdin)
+		return "", nil
+	}
```
g-linville, line 83: This makes the prompt look a little bit better if all we want to do is display a message to the user instead of getting information from the user.

pkg/runner/runner.go
```diff
@@ -681,7 +681,7 @@ func streamProgress(callCtx *engine.Context, monitor Monitor) (chan<- types.Comp
 					CallContext:      callCtx.GetCallContext(),
 					Type:             EventTypeCallProgress,
 					ChatCompletionID: status.CompletionID,
-					Content:          message.String(),
+					Content:          getEventContent(message.String(), *callCtx),
```
g-linville, line 684: This is to prevent credential tool output from leaking to the terminal.

pkg/types/tool.go
```diff
@@ -255,10 +255,10 @@ func ParseCredentialArgs(toolName string, input string) (string, string, map[str
...
 	inputMap := make(map[string]any)
 	if input != "" {
-		err := json.Unmarshal([]byte(input), &inputMap)
-		if err != nil {
-			return "", "", nil, fmt.Errorf("failed to unmarshal input: %w", err)
-		}
+		// Sometimes this function can be called with input that is not a JSON string.
+		// This typically happens during chat mode.
+		// That's why we ignore the error if this fails to unmarshal.
+		_ = json.Unmarshal([]byte(input), &inputMap)
```
StrongMonkey, line 261: Do you know that why this method is called twiced with different inputs?

StrongMonkey (approved):

pkg/prompt/prompt.go
```diff
@@ -75,6 +75,13 @@ func SysPrompt(ctx context.Context, envs []string, input string, _ chan<- string
 func sysPrompt(ctx context.Context, req types.Prompt) (_ string, err error) {
 	defer context2.GetPauseFuncFromCtx(ctx)()()
 
+	if req.Message != "" && len(req.Fields) == 1 && strings.TrimSpace(req.Fields[0]) == "" {
+		_, _ = fmt.Fprintln(os.Stderr, req.Message)
+		_, _ = fmt.Fprintln(os.Stderr, "Press enter to continue...")
+		_, _ = fmt.Fscanln(os.Stdin)
+		return "", nil
```
thedadams, line 82: Is it worth doing something like this?
```suggestion
		var errs []error
		 _, err := fmt.Fprintln(os.Stderr, req.Message)
		errs = append(errs, err)
		_, err = fmt.Fprintln(os.Stderr, "Press enter to continue...")
		errs = append(errs, err)
		_, err = fmt.Fscanln(os.Stdin)
		errs = append(errs, err)
		return "", errors.Join(errs...)
```

thedadams (approved):

pkg/prompt/prompt.go
g-linville: I suppose so. I have never seen any of these fail, but it's worth reporting if they do

PR #541, opened 2024-06-24
fix: bump go-gptscript to fix default model auth tui prompt

njhale:
Bump go-gptscript to include a fix that corrects the name of the default model environment variable passed to the sdk server. This prevents the TUI from prompting for an OpenAI key when using local models that don't require it.

Depends on https://github.com/gptscript-ai/go-gptscript/pull/34


go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/gptscript-ai/gptscript
 
 go 1.22.3
 
+replace github.com/gptscript-ai/go-gptscript => ../go-gptscript
```
njhale, line 5: note to self: drop this replace directive before merging

pkg/cli/gptscript.go
```diff
@@ -84,7 +84,7 @@ func New() *cobra.Command {
 		&Parse{},
 		&Fmt{},
 		&SDKServer{
-			GPTScript: root,
+			GPTScript: *root,
```
StrongMonkey, line 87: Is this change required?

njhale: required. part of the problem is that our cli builder package doesn't properly detect the option field when it's a pointer.

StrongMonkey (approved):

g-linville (approved):

thedadams (approved):

iwilltry42 (approved):

PR #546, opened 2024-06-25
fix: bump cmd module to allow anonymous struct pointers to propagate args

njhale:
Fixes a regression caused by switching from an embedded config pointer to a
struct value.

Also update acorn-io modules to their new gptscript-ai org names.


thedadams (approved):

njhale:
Going to hold off until `github.com/acorn-io/cmd` gets moved to `gptscript.com/gptscript-ai/cmd`.

thedadams (approved):

PR #554, opened 2024-06-25
enhance: add support for gptscript.env file in workspace

cloudnautique:
To add environment variables to a sys.exec call, the user can populate a gptscript.env file with a list of environment variables per line in the format:

```
FOO=bar
```

This file can be precreated or modified by gptscripts.

### DIFF
--- a/pkg/builtin/builtin.go
+++ b/pkg/builtin/builtin.go
@@ -336,6 +336,11 @@ func SysExec(_ context.Context, env []string, input string, progress chan<- stri
 		}
 		combined = io.MultiWriter(&out, &pw)
 	)
+
+	if envvars, err := getWorkspaceEnvFileContents(env); err == nil {
+		env = append(env, envvars...)
+	}
+
 	cmd.Env = env
 	cmd.Dir = params.Directory
 	cmd.Stdout = combined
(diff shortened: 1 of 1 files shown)


pkg/builtin/builtin.go
```diff
@@ -355,6 +360,36 @@ func (pw *progressWriter) Write(p []byte) (n int, err error) {
...
+		log.Debugf("The file %s does not exist", file)
+		return envContents, nil
+	} else if err != nil {
+		log.Debugf("Failed to read file %s: %v", file, err.Error())
+		return envContents, nil
+	}
+
+	envContents = append(envContents, strings.Split(string(data), "\n")...)
+
+	return envContents, nil
```
thedadams, line 390: nit: the `envContent` variable is not needed. You could do this:
```suggestion
	return strings.Split(string(data), "\n"), nil
```

thedadams: Also, do we want to filter out stuff that isn't in the form `FOO=bar`?

thedadams (approved):

njhale (approved):

thedadams (approved):

PR #555, opened 2024-06-25
docs: update context docs

cjellick:
Signed-off-by: Craig Jellick <craig@acorn.io>

### DIFF
--- a/docs/docs/03-tools/05-context.md
+++ b/docs/docs/03-tools/05-context.md
@@ -1,55 +1,50 @@
 # Context
 
-GPTScript provides a mechanism to share prompt information across many tools using the tool parameter `context`. It is used to provide additional information to the calling tool on when to use a specific tool by prepending the `context` to the instruction of the calling tool.
+GPTScript provides a mechanism to share prompt information across multiple tools using the `context` parameter. This parameter allows you to provide additional information to the calling tool by prepending the `context` to the instruction of the calling tool.
 
-- Context can point to a static text or a GPTScript.
-- Context tools are just regular GPTScript tools, and any valid gptscript field can be used.
+- Context can point to static text or a GPTScript.
+- Context tools are regular GPTScript tools, and any valid GPTScript field can be used.
 - Exported tools from a context tool are made available to the calling tool.
-- When context points to a GPTScript tool, output from the context tool gets prepended to the instruction of the calling tool.
+- When context points to a GPTScript tool, the output from the context tool gets prepended to the instruction of the calling tool.
 
-## Writing a Context Provider Tool as static text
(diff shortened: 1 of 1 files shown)


docs/docs/03-tools/05-context.md
```diff
@@ -59,24 +54,21 @@ context: ./my-search-context-tool.gpt
...
 What are some of the most popular tourist destinations in Scotland, and how many people visit them each year?
-
 ```
 
-When you run this script, GPTScript will use the output from the context tool and add it to the user message along with the 
-existing prompt in this tool to provide additional context to LLM.
+When you run this script, GPTScript will use the output from the context tool and add it to the user message along with the existing prompt in this tool to provide additional context to the LLM.
 
-## Context Provider Tool with args
+## Context Provider Tool with Args
```
njhale, line 61: Is the official verbiage "Args" or "Params"?

docs/docs/03-tools/05-context.md
```diff
@@ -87,11 +79,50 @@ context: ./context_with_arg.gpt with ${search} as search_tool
...
+name: Kevin
+context: ./context
+
+...
 ```
+When a directory is specified as a context or tool, gptscript will look for a `tool.gpt` in that directory and use it.
+
+The above example context tool has several interesting features:
+
+The `share context` directive is used to include additional context from other tools or files into the tool that referenced this context. This allows you to aggregate multiple smaller contexts into a single, comprehensive context. Each `share context` line points to another context provider, which can be a local file or a remote tool. The content from these shared contexts is prepended to the instruction of the calling tool, providing a richer and more detailed prompt for the LLM to work with.
```
njhale, line 124: We call preamble fields "directives" here and "tool parameters" elsewhere. Is it worth standardizing the verbiage across the docs?

cjellick: Would be nice to standardize...ive also used the word stanza. "directive" here was actually a new one generated by gpt 😅 

I do not like tool parameters because parameter is one of the parameters.

cjellick: im actually going to be adding more commits to this with other changes (with the goal of standardizing some other stuff).

njhale: > I do not like tool parameters because parameter is one of the parameters

hard concur

> im actually going to be adding more commits to this with other changes (with the goal of standardizing some other stuff)

sweet, thanks

njhale: Just to throw my vote in the hat, I think "tool field", "tool preamble field", or "tool definition field" are some descriptive options that are different _enough_ from "tool parameters" to prevent confusion.

PR #561, opened 2024-06-26
feat: build credential helpers just in time

thedadams:
Instead of building the credential helpers on startup, this change builds them when they are needed. This speeds up start time for the SDK.


g-linville (approved):

njhale (approved):
tested this out, works great! LGTM

Issue #509, opened 2024-06-18
Prompted for entering API keys when using local models which require no keys. 


sangee2004:
gptscript version v0.8.2+1b5e068f

**Steps to reproduce the problem:**
1. Host a LLM model locally 

2. Execute gptscript using this model.

3. User is prompted to enter API keys for the model which is incorrect.

```
 gptscript --disable-cache --default-model 'Llama-3-8b-function-calling-alpha-v1.gguf from http://localhost:1234/v1' examples/bob.gpt
12:15:55 WARNING: Changing the default model can have unknown behavior for existing tools. Use the model field per tool instead.
12:15:55 started  [main]
? Please provide your API key for http://localhost:1234/v1 
12:15:58 Saved API key as credential http://localhost:1234/v1
12:15:58 sent     [main]
12:16:00 ended    [main] [output=starttoolcall{\"name\": \"bob\", \"arguments\": \"{\\\"question\\\":\\\"How are you doing?\\\"}\"}endtoolcall]
12:16:00 usage    [total=220] [prompt=193] [completion=27]

OUTPUT:

starttoolcall{"name": "bob", "arguments": "{\"question\":\"How are you doing?\"}"}endtoolcall

```

4. This behavior is also seen when listing models

```
% gptscript --list-models http://localhost:1234/v1 
? Please provide your API key for http://localhost:1234/v1 
11:40:22 Saved API key as credential http://localhost:1234/v1
```

5. Not entering anything when prompted for credentials  (from step 3 and step 4), results in a credential entry being created after which I am not able to list credentials.

```
% gptscript credentials                           
2024/06/18 11:41:58 failed to list credentials: unexpected end of JSON input
```

Deleting the credential created allows for existing credentials to be listed.

```
%gptscript credential delete http://localhost:1234/v1 

% gptscript credential                                 
braveSearchAPIKey
sys.openai
```

g-linville:
@sangee2004 can you re-test this one? Should be fixed.

sangee2004:
Tested with gptscript version - `v0.0.0-dev-a42c9e31-dirty`. This issue is not seen anymore.

PR #567, opened 2024-06-26
fix: don't prompt for creds for local models

g-linville:
for #509 

Details in comments.

### DIFF
--- a/pkg/credentials/credential.go
+++ b/pkg/credentials/credential.go
@@ -49,13 +49,15 @@ func (c Credential) toDockerAuthConfig() (types.AuthConfig, error) {
 
 func credentialFromDockerAuthConfig(authCfg types.AuthConfig) (Credential, error) {
 	var cred Credential
-	if err := json.Unmarshal([]byte(authCfg.Password), &cred); err != nil || len(cred.Env) == 0 {
-		// Legacy: try unmarshalling into just an env map
-		var env map[string]string
-		if err := json.Unmarshal([]byte(authCfg.Password), &env); err != nil {
-			return Credential{}, err
+	if authCfg.Password != "" {
+		if err := json.Unmarshal([]byte(authCfg.Password), &cred); err != nil || len(cred.Env) == 0 {
+			// Legacy: try unmarshalling into just an env map
+			var env map[string]string
+			if err := json.Unmarshal([]byte(authCfg.Password), &env); err != nil {
+				return Credential{}, err
+			}
+			cred.Env = env
 		}
-		cred.Env = env
 	}
 
 	// We used to hardcode the username as "gptscript" before CredentialType was introduced, so
--- a/pkg/credentials/helper.go
+++ b/pkg/credentials/helper.go
@@ -2,6 +2,7 @@ package credentials
 
 import (
 	"errors"
+	"net/url"
 	"regexp"
 	"strings"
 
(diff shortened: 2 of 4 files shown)


pkg/credentials/credential.go
```diff
@@ -49,13 +49,15 @@ func (c Credential) toDockerAuthConfig() (types.AuthConfig, error) {
 
 func credentialFromDockerAuthConfig(authCfg types.AuthConfig) (Credential, error) {
 	var cred Credential
-	if err := json.Unmarshal([]byte(authCfg.Password), &cred); err != nil || len(cred.Env) == 0 {
-		// Legacy: try unmarshalling into just an env map
-		var env map[string]string
-		if err := json.Unmarshal([]byte(authCfg.Password), &env); err != nil {
-			return Credential{}, err
+	if authCfg.Password != "" {
```
g-linville, line 52: This fixes a bug where we would try to unmarshal an empty string.

pkg/credentials/helper.go
```diff
@@ -71,9 +72,18 @@ func (h *HelperStore) GetAll() (map[string]types.AuthConfig, error) {
...
+				if err != nil {
+					return nil, err
+				}
+
+				// Save the path so we can put it back after removing it.
+				path := toolURL.Path
+				toolURL.Path = ""
+
+				toolName = toolURL.String() + ":" + possiblePortNumber + path
 				ctx = strings.TrimSuffix(ctx, ":"+possiblePortNumber)
```
g-linville, line 87: This fixes a bug where we were putting the port after the path.
I.e., the string we're trying to fix is `http://localhost/v1///default:5555` and we used to make it become `http://localhost/v1:5555` rather than `http://localhost:5555/v1`.

pkg/remote/remote.go
```diff
@@ -108,7 +108,7 @@ func (c *Client) clientFromURL(ctx context.Context, apiURL string) (*openai.Clie
 	env := "GPTSCRIPT_PROVIDER_" + env2.ToEnvLike(parsed.Hostname()) + "_API_KEY"
 	key := os.Getenv(env)
 
-	if key == "" {
+	if key == "" && !isLocalhost(apiURL) {
```
g-linville, line 111: Here, we no longer look for the credential in the store if the model is running locally. Not sure whether we care to support IPv6 (`::1`) here, but I can add that if we do.

pkg/types/tool.go
```diff
@@ -283,6 +284,10 @@ func ParseCredentialArgs(toolName string, input string) (string, string, map[str
 		fields = fields[2:]
 	}
 
+	if alias != "" && !isAlphaNumeric(alias) {
+		return "", "", nil, fmt.Errorf("credential alias must be alphanumeric")
+	}
```
g-linville, line 289: I decided to add a constraint that a credential alias needs to be alphanumeric. This is to avoid interfering with credentials for model providers and such.

pkg/types/tool.go
```diff
@@ -780,3 +785,7 @@ func FirstSet[T comparable](in ...T) (result T) {
 	}
 	return
 }
+
+func isAlphaNumeric(s string) bool {
+	return regexp.MustCompile(`^[a-zA-Z0-9_]+$`).MatchString(s)
```
njhale, line 790: does this also need to include `.` to support common aliases in use today? e.g. `sys.openai` https://github.com/gptscript-ai/dalle-image-generation/blob/763a867b4e8ae0529782b74c1bcb08ecb197b2c8/tool.gpt#L2

g-linville: Thanks...I am very dumb for not thinking of this, lol

njhale: heh, the only reason I thought of it is because I was just looking at it lol

njhale (approved):

thedadams (approved):

PR #572, opened 2024-06-27
fix: only check cred helper repo once per 24h

g-linville:
We used to check the GH repo information for the credential helpers repo every time a cred store operation (get, list, etc.) was called. This is way too many lookups and hits the rate limit. This PR adds a file to hold a timestamp for the last time the repo was checked, and only check again if at least 24 hours have passed.

### DIFF
--- a/pkg/repos/get.go
+++ b/pkg/repos/get.go
@@ -114,12 +114,13 @@ func (m *Manager) deferredSetUpCredentialHelpers(ctx context.Context, cliCfg *co
 	defer locker.Unlock("gptscript-credential-helpers")
 
 	// Load the last-checked file to make sure we haven't checked the repo in the last 24 hours.
+	now := time.Now()
 	lastChecked, err := os.ReadFile(m.credHelperDirs.LastCheckedFile)
 	if err == nil {
-		if t, err := time.Parse(time.RFC3339, strings.TrimSpace(string(lastChecked))); err == nil && time.Since(t) < 24*time.Hour {
+		if t, err := time.Parse(time.RFC3339, strings.TrimSpace(string(lastChecked))); err == nil && now.Sub(t) < 24*time.Hour {
 			// Make sure the binary still exists, and if it does, return.
 			if _, err := os.Stat(filepath.Join(m.credHelperDirs.BinDir, "gptscript-credential-"+helperName+suffix)); err == nil {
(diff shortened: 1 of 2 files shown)


pkg/repos/get.go
```diff
@@ -112,11 +113,29 @@ func (m *Manager) deferredSetUpCredentialHelpers(ctx context.Context, cliCfg *co
 	locker.Lock("gptscript-credential-helpers")
 	defer locker.Unlock("gptscript-credential-helpers")
 
+	// Load the last-checked file to make sure we haven't checked the repo in the last 24 hours.
+	lastChecked, err := os.ReadFile(m.credHelperDirs.LastCheckedFile)
+	if err == nil {
+		if t, err := time.Parse(time.RFC3339, strings.TrimSpace(string(lastChecked))); err == nil && time.Since(t) < 24*time.Hour {
+			// Make sure the binary still exists, and if it does, return.
+			if _, err := os.Stat(filepath.Join(m.credHelperDirs.BinDir, "gptscript-credential-"+helperName+suffix)); err == nil {
+				log.Debugf("Not checking for new version of credential helper %s, last checked %v", helperName, t)
```
njhale, line 122: nit:

```suggestion
				log.Debugf("Credential helper %s up-to-date as of %v, checking for updates after %v", helperName, t, lastChecked.Add(24*time.Hour))
```

pkg/repos/get.go
```diff
@@ -112,11 +113,29 @@ func (m *Manager) deferredSetUpCredentialHelpers(ctx context.Context, cliCfg *co
...
+	}
+
+	// Load the credential helpers repo information.
 	_, repo, _, err := github.Load(ctx, nil, credentialHelpersRepo)
 	if err != nil {
 		return err
 	}
 
+	// Update the last-checked file.
+	if err := os.WriteFile(m.credHelperDirs.LastCheckedFile, []byte(time.Now().Format(time.RFC3339)), 0644); err != nil {
```
njhale, line 135: super-nit: If I wanted to be really pedantic, I'd recommend storing `time.Now()` in a variable, when you declare `lastChecked` above, and base all the comparisons off of that.

thedadams (approved):

pkg/repos/get.go
g-linville: Fair enough lol

PR #578, opened 2024-06-27
fix: use GH token for VCS lookups

g-linville:
With this change, we now use the GitHub token for VCS lookups when resolving tools. This can help avoid rate limits.

I also fix the Claude smoke test in this PR, but it runs the old version against this PR, which is why it is still failing.

### DIFF
--- a/.github/workflows/smoke.yaml
+++ b/.github/workflows/smoke.yaml
@@ -139,7 +139,7 @@ jobs:
           go-version: "1.21"
       - env:
           OPENAI_API_KEY: ${{ secrets.SMOKE_OPENAI_API_KEY }}
-          GPTSCRIPT_DEFAULT_MODEL: claude-3-opus-20240229 from github.com/gptscript-ai/claude3-anthropic-provider@main
+          GPTSCRIPT_DEFAULT_MODEL: claude-3-opus-20240229 from github.com/gptscript-ai/claude3-anthropic-provider
           ANTHROPIC_API_KEY: ${{ secrets.SMOKE_ANTHROPIC_API_KEY }}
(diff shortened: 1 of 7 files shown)


.github/workflows/smoke.yaml
```diff
@@ -139,7 +139,7 @@ jobs:
           go-version: "1.21"
       - env:
           OPENAI_API_KEY: ${{ secrets.SMOKE_OPENAI_API_KEY }}
-          GPTSCRIPT_DEFAULT_MODEL: claude-3-opus-20240229 from github.com/gptscript-ai/claude3-anthropic-provider@tool-beta
+          GPTSCRIPT_DEFAULT_MODEL: claude-3-opus-20240229 from github.com/gptscript-ai/claude3-anthropic-provider
```
njhale, line 142: I think you'll need to do a credential override for this to work now that https://github.com/gptscript-ai/claude3-anthropic-provider/commit/cc53b05ea7649b8ee1a3020f4dc4a5dd11fd7b68 merged, right?

g-linville: Oh good point, I guess I will

njhale (approved):

thedadams:
@njhale and I just fixed the claude smoke tests (was a problem in the `tool-beta` branch over there). So I re-ran and they should pass.

njhale:
@g-linville Also, just noting that members of the `gptscript-ai` org don't need to add the `run-smoke` label for smoke tests to run for their PRs.

LMK if this isn't the case for you and I can look into it.

g-linville:
Thanks. Yeah I didn't need to add the label but I added it at some point to see if it would change anything while I was troubleshooting, lol.

PR #580, opened 2024-06-27
fix: remove requirement that cred aliases be alphanumeric

g-linville:
This requirement would probably cause us more problems than it's worth (like weird edge cases in tool generation for OpenAPI), so it's probably best that we just remove it.

### DIFF
--- a/pkg/types/tool.go
+++ b/pkg/types/tool.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"path/filepath"
-	"regexp"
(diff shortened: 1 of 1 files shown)


g-linville:
The Claude smoke test will be broken because the provider for it is (temporarily) broken. This will be fixed shortly.

cjellick (approved):

thedadams (approved):

StrongMonkey (approved):

drpebcak (approved):

PR #586, opened 2024-06-28
fix: create the cache directory before writing last-checked file

thedadams:
If the cache directory is gone, then, currently, writing the last-checked file fails. This change ensures that the directory exists before trying to write the file.


njhale (approved):

g-linville (approved):

g-linville:
Whoops! Thanks for catching this

Issue #321, opened 2024-05-06
OpenAPI generate function names if operationIds are unavailable


cloudnautique:
If operationIds are not provided in the OpenAPI spec, the parser should autogenerate based on "method" and "path" or similar so a consumer an use the spec without having to download and modify it.

g-linville:
This one is done @sangee2004

sangee2004:
gptscript version v0.9.2+0f32a9d8

Able to work with openapi yamls even when there are no operationIds present.

Tested with following openapi yaml which does not have operationIds for list and get methods:
```
# Adapted from the official example at https://github.com/OAI/OpenAPI-Specification/blob/66fe9db36115bbf5425892aaaac6dba5e3c5df59/examples/v3.0/petstore.yaml
openapi: "3.0.0"
info:
  version: 1.0.0
  title: Swagger Petstore
  license:
    name: MIT
paths:
  /pets:
    get:
      summary: List all pets
      tags:
        - pets
      parameters:
        - name: limit
          in: query
          description: How many items to return at one time (max 100)
          required: false
          schema:
            type: integer
            maximum: 100
            format: int32
      responses:
        '200':
          description: A paged array of pets
          headers:
            x-next:
              description: A link to the next page of responses
              schema:
                type: string
          content:
            application/json:    
              schema:
                $ref: "#/components/schemas/Pets"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
  /pets/{petId}:
    get:
      summary: Info for a specific pet
      tags:
        - pets
      parameters:
        - name: petId
          in: path
          required: true
          description: The id of the pet to retrieve
          schema:
            type: string
      responses:
        '200':
          description: Expected response to a valid request
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Pet"
        default:
          description: unexpected error
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/Error"
components:
  schemas:
    Pet:
      type: object
      required:
        - id
        - name
      properties:
        id:
          type: integer
          format: int64
        name:
          type: string
        tag:
          type: string
    Pets:
      type: array
      maxItems: 100
      items:
        $ref: "#/components/schemas/Pet"
    Error:
      type: object
      required:
        - code
        - message
      properties:
        code:
          type: integer
          format: int32
        message:
          type: string
```

Able to see the following script work successfully:
```
Tools: http://localhost:8080/openapi

[...]

PR #601, opened 2024-07-02
enhance: generate name for OpenAPI tools when operation ID is blank

g-linville:
for #321

### DIFF
--- a/pkg/loader/openapi.go
+++ b/pkg/loader/openapi.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"regexp"
 	"slices"
 	"sort"
 	"strings"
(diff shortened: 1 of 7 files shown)


pkg/loader/openapi.go
```diff
@@ -115,6 +116,13 @@ func getOpenAPITools(t *openapi3.T, defaultHost string) ([]types.Tool, error) {
 				toolDesc = toolDesc[:1024]
 			}
 
+			toolName := operation.OperationID
+			if toolName == "" {
+				// When there is no operation ID, we use the method + path as the tool name and remove all characters
+				// except letters, numbers, underscores, and hyphens.
+				toolName = regexp.MustCompile(`[^a-zA-Z0-9_-]+`).ReplaceAllString(strings.ToLower(method)+strings.ReplaceAll(pathString, "/", "_"), "")
+			}
+
```
njhale, line 125: nit: Not a blocker, but some autogold unit tests for `getOpenAPITools(...)` would make me much more comfortable with changes like this, and at first glance it doesn't look like they would be too much extra work to add.

njhale (approved):

pkg/loader/openapi.go
```diff
@@ -115,6 +116,13 @@ func getOpenAPITools(t *openapi3.T, defaultHost string) ([]types.Tool, error) {
 				toolDesc = toolDesc[:1024]
 			}
 
+			toolName := operation.OperationID
+			if toolName == "" {
+				// When there is no operation ID, we use the method + path as the tool name and remove all characters
+				// except letters, numbers, underscores, and hyphens.
+				toolName = regexp.MustCompile(`[^a-zA-Z0-9_-]+`).ReplaceAllString(strings.ToLower(method)+strings.ReplaceAll(pathString, "/", "_"), "")
```
thedadams, line 123: nit: in general, it's bad practice to have things that could panic in code. It is unlikely that this will panic, but I wanted to call this out.

Typically, creating a package variable that is initialized on package initialization (i.e. on startup) is preferable.

thedadams (approved):

pkg/loader/openapi.go
g-linville: Great idea

njhale (approved):

pkg/loader/openapi.go
g-linville: Added these

thedadams (approved):

PR #611, opened 2024-07-03
Revert "Fix: Try to send interupt to sys.daemon to allow cleanup"

StrongMonkey:
This reverts commit 1104843ccc7c571568756e8468d7607d546fb467.

Causing issue in other LLM provider, reverting.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -5,7 +5,6 @@ import (
 	"io"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


drpebcak (approved):

njhale (approved):
I've confirmed that, at minimum, the regression affects both the claude and mistral providers. I suspect it will cause problems for the other providers as well.

I've also verified that this revert fixes the problem for both providers.

PR #634, opened 2024-07-15
enhance: share credential

g-linville:
This adds a new tool parameter, `Share Credentials` (with several aliases), which is essentially the same thing as `Share Tools` but specifically for credential tools.

I also added a new integration test that tests credential "scopes" to make sure that in a tree of tools that define and share credentials, each tool only has access to credentials that it defines and credentials shared by its immediate children in the tree.

### DIFF
--- a/pkg/parser/parser.go
+++ b/pkg/parser/parser.go
@@ -148,7 +148,7 @@ func isParam(line string, tool *types.Tool) (_ bool, err error) {
 		}
 	case "credentials", "creds", "credential", "cred":
 		tool.Parameters.Credentials = append(tool.Parameters.Credentials, value)
-	case "exportcredentials", "exportcreds", "exportcredential", "exportcred", "sharecredentials", "sharecreds", "sharecredential", "sharecred":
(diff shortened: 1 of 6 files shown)


pkg/parser/parser.go
```diff
@@ -148,6 +148,8 @@ func isParam(line string, tool *types.Tool) (_ bool, err error) {
 		}
 	case "credentials", "creds", "credential", "cred":
 		tool.Parameters.Credentials = append(tool.Parameters.Credentials, value)
+	case "exportcredentials", "exportcreds", "exportcredential", "exportcred", "sharecredentials", "sharecreds", "sharecredential", "sharecred":
```
ibuildthecloud, line 151: don't support the "export" keyword. For new things we are only adding "share" since we want everything to switch to share.

g-linville: fixed

g-linville: Well I pushed another commit at least, but it isn't showing up in the PR 🤔

thedadams (approved):

StrongMonkey (approved):
💯 on the test added

PR #646, opened 2024-07-17
Docs - Add faq for github rate limit

sangee2004:
Added workaround that can be done when gptscript execution results in github rate limit errors in FAQ section.

### DIFF
--- a/docs/docs/09-faqs.md
+++ b/docs/docs/09-faqs.md
@@ -53,3 +53,11 @@ This tells the LLM (by way of a [system message](https://platform.openai.com/doc
 
 This context also automatically shares the `sys.ls`, `sys.read`, and `sys.write` tools with the tool that is using it as a context. This is because if a tool intends to interact with the workspace, it minimally needs these tools.
 
+### I'm hitting GitHub's rate limit for unauthenticated requests when using GPTScript.
+
+By default, GPTScript makes unauthenticated requests to GitHub when pulling tools. Since GitHub's rate limits for unauthenticated requests are fairly low, running into them when developing with GPTScript is a common issue. To avoid this, you can get GPTScript to make authenticated requests -- which have higher rate limits -- by setting the `GITHUB_AUTH_TOKEN` environment variable to your github account's PAT (Personal Access Token).
+If you're already authenticated with the `gh` CLI, you can use its token by running:
+
+```bash
+export GITHUB_AUTH_TOKEN="$(gh auth token)"
+```


docs/docs/09-faqs.md
```diff
@@ -53,3 +53,6 @@ This tells the LLM (by way of a [system message](https://platform.openai.com/doc
 
 This context also automatically shares the `sys.ls`, `sys.read`, and `sys.write` tools with the tool that is using it as a context. This is because if a tool intends to interact with the workspace, it minimally needs these tools.
 
+### I am rate limted by GitHub when executing GPTScripts. What can be done to workaround this?
```
njhale, line 56:
```suggestion
### I'm hitting GitHub's rate limit for unauthenticated requests when using GPTScript.

By default, GPTScript makes unauthenticated requests to GitHub when pulling tools. Since GitHub's rate limits for unauthenticated requests are fairly low, running into them when developing with GPTScript is a common issue. To avoid this, you can get GPTScript to make authenticated requests -- which have higher rate limits -- by setting the `GITHUB_AUTH_TOKEN` environment variable.

If you're already authenticated with the `gh` CLI, you can use its token by running:

```bash
export GITHUB_AUTH_TOKEN="$(gh auth token)"
```

```

**Review by njhale — COMMENTED** [review]
Thanks for adding this @sangee2004!

Left a suggestion. LMK what you think.

File: docs/docs/09-faqs.md (hunk 1 above)
Comment (drpebcak) [reply docs/docs/09-faqs.md:56]: I would just add a mention to the fact that this should be a PAT (Personal Access Token). If people want to set it from the github cli tool that is fine, but being more explicit about it for people not familiar with that tool would be better.

**Comment by sangee2004** [comment]
@njhale @drpebcak I have made changes as per your suggestions.

File: docs/docs/09-faqs.md [hunk 2]
```diff
@@ -53,3 +53,10 @@ This tells the LLM (by way of a [system message](https://platform.openai.com/doc
 
 This context also automatically shares the `sys.ls`, `sys.read`, and `sys.write` tools with the tool that is using it as a context. This is because if a tool intends to interact with the workspace, it minimally needs these tools.
 
+### I'm hitting GitHub's rate limit for unauthenticated requests when using GPTScript.
+
+By default, GPTScript makes unauthenticated requests to GitHub when pulling tools. Since GitHub's rate limits for unauthenticated requests are fairly low, running into them when developing with GPTScript is a common issue. To avoid this, you can get GPTScript to make authenticated requests -- which have higher rate limits -- by setting the `GITHUB_AUTH_TOKEN` environment variable to your github account's PAT (Personal Access Token).
+If you're already authenticated with the `gh` CLI, you can use its token by running:
+```bash
+export GITHUB_AUTH_TOKEN="$(gh auth token)"
+```
```
njhale, line 62:
```suggestion
If you're already authenticated with the `gh` CLI, you can use its token by running:

```bash
export GITHUB_AUTH_TOKEN="$(gh auth token)"
```
```

njhale (approved):
Just one formatting nit, but other than that, LGTM!

drpebcak (approved):
LGTM as well, pending that nit

njhale (approved):

PR #652, opened 2024-07-22
fix: use default config in embedded SDK server

thedadams:


pkg/sdkserver/server.go
```diff
@@ -56,6 +56,8 @@ func Run(ctx context.Context, opts Options) error {
 // EmbeddedStart allows running the server as an embedded process that may use Stdin for input.
 // It returns the address the server is listening on.
 func EmbeddedStart(ctx context.Context, opts Options) (string, error) {
```
ibuildthecloud, line 58: why you no like my variadic approach :(  I love doing `opts ...Options` now. It's so nicely composable.

thedadams: I can update that.

thedadams: Done.

PR #661, opened 2024-07-25
fix: Append all credentials for OpenAPI security infos

johnrengelman:
When using an OpenAPI definition with multiple security infos that should be AND'd together (per https://docs.gptscript.ai/tools/openapi#1-security-schemes), the tools credential injection with only include 1 of the environment variables instead of all of them. The result is non-deterministic on which envvar will be included to do Go's behavior with slices.
This changes the behavior to include credential injections for all schemes declared in the 1st security info block in the api definition.

### DIFF
--- a/pkg/loader/openapi.go
+++ b/pkg/loader/openapi.go
@@ -305,7 +305,7 @@ func getOpenAPITools(t *openapi3.T, defaultHost, source, targetToolName string) 
 					if err != nil {
 						return nil, fmt.Errorf("failed to parse operation server URL: %w", err)
 					}
-					tool.Credentials = info.GetCredentialToolStrings(operationServerURL.Hostname())
+					tool.Credentials = append(tool.Credentials, info.GetCredentialToolStrings(operationServerURL.Hostname())...)
 				}
 			}
 


johnrengelman:
Tried to add tests for this, but the current openApi loading tests don't handle tool linking and wasn't quite sure how to set that up in the test fixture. Got an error like so - 

```
failed linking listPets at : failed resolving github.com/gptscript-ai/credential as petstore.swagger.iobearerAuth with GPTSCRIPT_PETSTORE_SWAGGER_IO_BEARERAUTH as env and "Please provide a value for the GPTSCRIPT_PETSTORE_SWAGGER_IO_BEARERAUTH environment variable" as message and "bearer token" as field from : can not load tools path= name=github.com/gptscript-ai/credential
```

g-linville (approved):

pkg/loader/openapi.go
```diff
@@ -295,7 +295,9 @@ func getOpenAPITools(t *openapi3.T, defaultHost string) ([]types.Tool, error) {
 					if err != nil {
 						return nil, fmt.Errorf("failed to parse operation server URL: %w", err)
 					}
-					tool.Credentials = info.GetCredentialToolStrings(operationServerURL.Hostname())
+					for _, cred := range info.GetCredentialToolStrings(operationServerURL.Hostname()) {
+						tool.Credentials = append(tool.Credentials, cred)
+					}
```
g-linville, line 300: The linter wants to change it to this:

```suggestion
					tool.Credentials = append(tool.Credentials, info.GetCredentialToolStrings(operationServerURL.Hostname())...)
```

I tested it and it works the same.

g-linville:
Hey @johnrengelman , thanks for the contribution. Sorry it took me so long to get to this. It looks good to me but the linter is failing. If you could apply the suggestion I left, then I can merge this. Thanks!

johnrengelman:
Pushed this change from the linter.
I tried to look into the openai_revamp implementation to see if this same problem exists, but I wasn't able to follow where the credential envvars are injected in that (the tools are built up differently).

g-linville:
> Pushed this change from the linter.
> I tried to look into the openai_revamp implementation to see if this same problem exists, but I wasn't able to follow where the credential envvars are injected in that (the tools are built up differently).

Yeah I think we are planning on removing that. That was mostly just an experiment, which is why it is hidden behind an env var. So no worries. Thanks again for your contribution!

g-linville (approved):

PR #679, opened 2024-07-30
docs: Add faq for "Free Trial" openai accounts

sangee2004:
Add faq for "Free Trial" openai accounts

### DIFF
--- a/docs/docs/09-faqs.md
+++ b/docs/docs/09-faqs.md
@@ -61,3 +61,16 @@ If you're already authenticated with the `gh` CLI, you can use its token by runn
 ```bash
 export GITHUB_AUTH_TOKEN="$(gh auth token)"
 ```
+
+### I'm not able to get GPTScript to work with my OpenAI account which is in "Free Trial" 
+
+When executing GPTScript with OpenAI API key of an account in "Free Trial", user will be presented with the following error message: 
+```
+status code: 404, message: The model `gpt-4o` does not exist or you do not have access to it
+``` 
+By default, GPTScript uses `gpt-4o` model which is not available for users in "Free Trial" and this is expected behavior.
+
+User can choose to use other supported models like `gpt-4o-mini`,`gpt-3.5-turbo` by using `--default-model <Model Name>` when executing GPTscripts. But in case of "Free Trial" account with no credits, GPTScript execution will fail with following quota exceeded error messgae which is also the expected behavior since there are no credits available.
+```
+error, status code: 429, message: You exceeded your current quota, please check your plan and billing details. For more information on this error, read the docs: https://platform.openai.com/docs/guides/error-codes/api-errors.
+```


docs/docs/09-faqs.md
```diff
@@ -61,3 +61,16 @@ If you're already authenticated with the `gh` CLI, you can use its token by runn
 ```bash
 export GITHUB_AUTH_TOKEN="$(gh auth token)"
 ```
+
+### I'm not able to get GPTScript to work with my OpenAI account which is in "Free Trial" 
+
+When executing GPTScript with OpenAI API key of an account in "Free Trial", user will be presented with the following error message:
```
drpebcak, line 67:
```suggestion
When running GPTScript against an OpenAI account that is in "Free Trial" mode, you will be presented with the following error message:

```

docs/docs/09-faqs.md
```diff
@@ -61,3 +61,16 @@ If you're already authenticated with the `gh` CLI, you can use its token by runn
...
 export GITHUB_AUTH_TOKEN="$(gh auth token)"
 ```
+
+### I'm not able to get GPTScript to work with my OpenAI account which is in "Free Trial" 
+
+When executing GPTScript with OpenAI API key of an account in "Free Trial", user will be presented with the following error message: 
+```
+status code: 404, message: The model `gpt-4o` does not exist or you do not have access to it
+``` 
+By default, GPTScript uses `gpt-4o` model which is not available for users in "Free Trial" and this is expected behavior.
```
drpebcak, line 71:
```suggestion

GPTScript uses the `gpt-4o` model by default, which is not available for users in "Free Trial."
```

docs/docs/09-faqs.md
```diff
@@ -61,3 +61,16 @@ If you're already authenticated with the `gh` CLI, you can use its token by runn
...
+
+User can choose to use other supported models like `gpt-4o-mini`,`gpt-3.5-turbo` by using `--default-model <Model Name>` when executing GPTscripts. But in case of "Free Trial" account with no credits, GPTScript execution will fail with following quota exceeded error messgae which is also the expected behavior since there are no credits available.
```
drpebcak, line 73:
```suggestion
You can choose to use other supported models like `gpt-4o-mini` or `gpt-3.5-turbo` by using `--default-model <Model Name>` when executing scripts. If there are no available credits in the account, execution will fail with following quota exceeded error message until you add credits to your account.

```

docs/docs/09-faqs.md
```diff
@@ -61,3 +61,16 @@ If you're already authenticated with the `gh` CLI, you can use its token by runn
 ```bash
 export GITHUB_AUTH_TOKEN="$(gh auth token)"
 ```
+
+### I'm not able to get GPTScript to work with my OpenAI account which is in "Free Trial"
```
drpebcak, line 65: Ive added some suggestions to the text targeted at answering this exact question, but really there are sort of 2 scenarios that could lead to the '404 model not found' error. It might be worth splitting this into two sections, one that directly addresses the question of what to do if you get that error, and the other that addresses the rate limit error.

I think people will more frequently be asking questions about those errors they see directly from gptscript, where for now this is question is working backwards from an exact answer.

drpebcak (changes requested):

PR #688, opened 2024-07-31
feat: add support for default model provider in the TUI

thedadams:


pkg/sdkserver/types.go
```diff
@@ -63,7 +63,7 @@ type toolOrFileRequest struct {
 	Confirm              bool     `json:"confirm"`
 	Location             string   `json:"location,omitempty"`
 	ForceSequential      bool     `json:"forceSequential"`
-	DefaultModelProvider string   `json:"defaultModelProvider,omitempty"`
+	DefaultModelProvider string   `json:"DefaultModelProvider,omitempty"`
```
g-linville, line 66: Was this capitalization change intentional? Just want to make sure.

iwilltry42 (approved):

pkg/sdkserver/types.go
thedadams: Yes it was. Thanks for checking.

PR #692, opened 2024-08-01
feat: add basic bash support for windows

drpebcak:


ibuildthecloud (approved):

pkg/engine/cmd.go
```diff
@@ -203,10 +204,9 @@ var ignoreENV = map[string]struct{}{
 }
 
 func appendEnv(envs []string, k, v string) []string {
-	for _, k := range []string{k, env.ToEnvLike(k)} {
-		if _, ignore := ignoreENV[k]; !ignore {
-			envs = append(envs, k+"="+v)
-		}
+	//fmt.Printf("%s=%s\n", k, v)
```
thedadams, line 207: nit: remove

pkg/engine/cmd.go
```diff
@@ -203,10 +204,9 @@ var ignoreENV = map[string]struct{}{
...
+	if _, ignore := ignoreENV[k]; !ignore {
+		envs = append(envs, strings.ToUpper(k)+"="+v)
```
thedadams, line 209: Do we want to use `strings.ToUpper(env.ToEnvLike(k))` here to get right of `-` and `.`?

drpebcak: in my heart i hope that nobody is trying to create env vars like that, but you are right it would not hurt to do!

thedadams (approved):

PR #708, opened 2024-08-05
enhance: remove --ui option

njhale:
Remove the ability to start the UI tool from gptscript. This drops the
`--ui` flag from the gptscript command.

Note: Let's hold off on merging this until electron support is merged into https://github.com/gptscript-ai/ui.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -12,13 +12,6 @@ jobs:
   release-tag:
     runs-on: ubuntu-22.04
     steps:
-      - name: trigger ui repo tag workflow
-        uses: peter-evans/repository-dispatch@v3
-        with:
-          token: ${{ secrets.DISPATCH_PAT }}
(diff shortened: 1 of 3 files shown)


thedadams (approved):
These changes look good. Should we also remove the workflow that tags the UI? Last I heard, the UI electron app was going to be on a slightly different release cycle.

njhale:
@thedadams 

>  Should we also remove the workflow that tags the UI? Last I heard, the UI electron app was going to be on a slightly different release cycle

I think that's a good idea since there's currently at least one manual step -- bumping the `version` field in the package.json file -- to releasing. As it stands, we'll end up inadvertently updating the published builds for the previous release of the UI when gptscript releases.

I'll drop the workflow now.

njhale:
done

thedadams (approved):

tylerslaton (approved):

PR #728, opened 2024-08-06
fix: windows absolute path logic

drpebcak:
this fixes an issue seen at least in clio, with the ability to load external agents.

### DIFF
--- a/pkg/loader/loader.go
+++ b/pkg/loader/loader.go
@@ -68,10 +68,8 @@ func openFile(path string) (io.ReadCloser, bool, error) {
 }
 
 func loadLocal(base *source, name string) (*source, bool, error) {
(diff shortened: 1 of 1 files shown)


drpebcak:
This change was originally suggested by Darren, but I'm not sure he intended it to be the actual fix since it changes the behavior mentioned in the comment. Does anyone have an alternate suggestion?

ibuildthecloud:
The comment is correct, but in the wrong place.  We want to use path.Join and not filepath.Join. As some point the IsAbs check was added and the comment should have gone inside the if.  We want the comment because it still holds true. IsAbs does not mutate the path but just checks for both cases.

ibuildthecloud (approved):

PR #755, opened 2024-08-09
fix: use a different shlex

thedadams:
The shlex splitting was ruining Windows paths. The new shlex is nicer to Windows.


njhale:
Looks like `shellwords` passes everything: https://github.com/gptscript-ai/gptscript/pull/756

PR #756, opened 2024-08-09
fix: use shellwords instead of shlex

njhale:
The shlex splitting was ruining Windows paths. The shellwords is nicer to Windows.

### DIFF
--- a/pkg/engine/cmd.go
+++ b/pkg/engine/cmd.go
@@ -17,7 +17,7 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/google/shlex"
+	"github.com/buildkite/shellwords"
 	"github.com/gptscript-ai/gptscript/pkg/counter"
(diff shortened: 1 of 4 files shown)


pkg/engine/cmd.go
```diff
@@ -17,7 +17,8 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/google/shlex"
+	// "github.com/anmitsu/go-shlex"
```
g-linville, line 20: nit: remove this and other comments

g-linville (approved):

pkg/engine/cmd.go
njhale: ah, thought I dropped these. Thanks for catching. done.

njhale:
Superseded by https://github.com/gptscript-ai/gptscript/pull/758

PR #799, opened 2024-08-15
fix: use default model provider if client is not found for model

thedadams:


njhale (approved):

thedadams:
Closing in favor of another approach.

PR #809, opened 2024-08-20
interp/embedded(yaegi): init embedded interpreter

rinor:
Experimenting with an embedded interpreter. Just a draft implementation to start a feasibility discussion and mostly to check how disruptive would the code be.

In case you previously discussed and discarded it as an option, feel free to drop and ignore it. 

---

- allows to pass a file to the embedded interpreter: `#!sys.yaegi /full_path_to/at-syntax.go`

```go
Name: At syntax
Description: Add the ability for users to use at (@) to direct message an agent
Type: context

Share Input Filter: At syntax filter

---
Name: At syntax filter
Description: Translates @name command into "Call tool name with command"

#!sys.yaegi /full_path_to/at-syntax.go
```

- or get it from the instructions

```go
Name: At syntax
Description: Add the ability for users to use at (@) to direct message an agent
Type: context

Share Input Filter: At syntax filter

---
Name: At syntax filter
Description: Translates @name command into "Call tool name with command"

#!sys.yaegi
package main

import (
	"bytes"
	"compress/gzip"
	"encoding/base64"
	"fmt"
	"io"
	"os"
	"strings"
)

func GetEnv(key, def string) string {
	v := os.Getenv(key)
	if v == "" {
		return def
	}

	if strings.HasPrefix(v, "{\"_gz\":\"") && strings.HasSuffix(v, "\"}") {
		data, err := base64.StdEncoding.DecodeString(v[8 : len(v)-2])
		if err != nil {
			return v
		}
		gz, err := gzip.NewReader(bytes.NewBuffer(data))
		if err != nil {
			return v
		}
		strBytes, err := io.ReadAll(gz)
		if err != nil {
			return v
		}
		return string(strBytes)
	}

	return v
}

func output(input string) string {
	input = strings.TrimSpace(input)
	if !strings.HasPrefix(input, "@") {
		return input
	}

	parts := strings.Fields(input[1:])
	if len(parts) == 1 {
		return fmt.Sprintf(`Call the tool "%s" with defaultPromptParameter as empty`, parts[0])
	}

	return fmt.Sprintf(`Call the tool "%s" with defaultPromptParameter as exactly "%s"`, parts[0], strings.Join(parts[1:], " "))
}

func main() {
	fmt.Println(output(GetEnv("INPUT", "")))
}

```


rinor:
check https://github.com/gptscript-ai/gptscript/pull/830#issuecomment-2318672673

PR #813, opened 2024-08-21
fix: remove config file location from the config

g-linville:
There was never a good reason to store it there anyway.

### DIFF
--- a/pkg/config/cliconfig.go
+++ b/pkg/config/cliconfig.go
@@ -52,14 +52,14 @@ func (a *AuthConfig) UnmarshalJSON(data []byte) error {
 }
 
 type CLIConfig struct {
-	Auths               map[string]AuthConfig `json:"auths,omitempty"`
-	CredentialsStore    string                `json:"credsStore,omitempty"`
-	GPTScriptConfigFile string                `json:"gptscriptConfig,omitempty"`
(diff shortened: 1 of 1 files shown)


cjellick (approved):

pkg/config/cliconfig.go
```diff
@@ -52,14 +52,14 @@ func (a *AuthConfig) UnmarshalJSON(data []byte) error {
...
-	GatewayURL          string                `json:"gatewayURL,omitempty"`
-	Integrations        map[string]string     `json:"integrations,omitempty"`
+	Auths            map[string]AuthConfig `json:"auths,omitempty"`
+	CredentialsStore string                `json:"credsStore,omitempty"`
+	GatewayURL       string                `json:"gatewayURL,omitempty"`
+	Integrations     map[string]string     `json:"integrations,omitempty"`
 
 	auths     map[string]types.AuthConfig
 	authsLock *sync.Mutex
+	location  string
```
thedadams, line 62: It doesn't look like this field is getting set anywhere?

thedadams (approved):

pkg/config/cliconfig.go
thedadams: It is. I just can't read.

PR #814, opened 2024-08-21
fix: surface the error message when we fail to find a release on GitHub

g-linville:


pkg/repos/runtimes/golang/golang.go
```diff
@@ -124,41 +126,39 @@ func getLatestRelease(tool types.Tool) (*release, bool) {
...
 	resp, err := client.Get(fmt.Sprintf("https://api.github.com/repos/%s/%s/tags", account, repo))
 	if err != nil {
-		// ignore error
-		return nil, false
+		return nil, false, err
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
-		return nil, false
+		return nil, false, fmt.Errorf("got status code %d from GitHub", resp.StatusCode)
```
thedadams, line 133: We shouldn't return an error if we get a 404.

pkg/repos/runtimes/golang/golang.go
```diff
@@ -124,41 +126,39 @@ func getLatestRelease(tool types.Tool) (*release, bool) {
...
 	resp, err = client.Get(fmt.Sprintf("https://github.com/%s/%s/releases/latest", account, repo))
 	if err != nil {
-		// ignore error
-		return nil, false
+		return nil, false, err
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusFound {
-		return nil, false
+		return nil, false, fmt.Errorf("got status code %d from GitHub", resp.StatusCode)
```
thedadams, line 156: Same.

PR #815, opened 2024-08-21
chore: add content length when handling json request bodies.

cloudnautique:
Explicitly setting the content length instead of allowing the Go default of transfer encoding chunked.
This change only applies to the application/json request body objects using openapi.


cloudnautique:
@drpebcak can you test this with some of your gptscript openapi integrations?

drpebcak (approved):

PR #817, opened 2024-08-23
Fix: Do not return full env map after downloading releases

StrongMonkey:
After downloading gptscript release, we return a full map of current environment variables and this will eventuall get saved in to done.file, which will be picked up and saved as environment variable on the next run. This is causing weird issues like https://github.com/gptscript-ai/desktop/issues/253. The original purpose of adding env variable to done.file is probably related to only append PATH, as we can see the logic here: https://github.com/gptscript-ai/gptscript/blob/56b286bd2b01fb1dd5b562f3d4c98d55147c4a20/pkg/repos/runtimes/golang/golang.go#L276. But this is not needed when downloading releases because it is always downloaded in known location.

### DIFF
--- a/pkg/repos/runtimes/golang/golang.go
+++ b/pkg/repos/runtimes/golang/golang.go
@@ -244,7 +244,7 @@ func getChecksum(ctx context.Context, rel *release, artifactName string) string 
 	return ""
 }
 
-func (r *Runtime) Binary(ctx context.Context, tool types.Tool, _, toolSource string, env []string) (bool, []string, error) {
+func (r *Runtime) Binary(ctx context.Context, tool types.Tool, _, toolSource string, _ []string) (bool, []string, error) {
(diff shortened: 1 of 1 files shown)


iwilltry42 (approved):
The go runtime is the only one even implementing the Binary() interface func, so I guess we could completely remove the env param, couldn't we?
Especially given your argument, that it doesn't make sense 😁

StrongMonkey:
yes, I don't actually want to change too much scope just in case though.

g-linville (approved):

thedadams (approved):

StrongMonkey:
>You dont think we're passing down anything else useful in these env vars?

I looked at other code and the only thing we are doing is to pass PATH environment so that the binary we built is baked into the PATH. Which i don't think we needed for downloading releases. So I think this change should be good and darren did not include this intentionally.

PR #818, opened 2024-08-23
interp/embedded(expr): init embedded expr-lang.org

rinor:
Experimenting with https://expr-lang.org/. Just a draft implementation to start a feasibility discussion and mostly to check where and how might be useful in practice. 

Right now I just converted some stuff from https://github.com/gptscript-ai/context :

**Note**: For testing, I'm using only builtin stuff [language-definition](https://github.com/expr-lang/expr/blob/master/docs/language-definition.md). Things can be improved by exposing additional custom data/functions/types.

- https://github.com/gptscript-ai/context/tree/main/at-syntax
```go
Name: At syntax
Description: Add the ability for users to use at (@) to direct message an agent
Type: context

Share Input Filter: At syntax filter

---
Name: At syntax filter
Description: Translates @name command into "Call tool name with command"

#!sys.interp.expr
// INPUT - variant
let input = INPUT == nil || len(INPUT) == 0 ? "" : trim(INPUT);
!(input startsWith '@') || len(input) < 2 ? input :
  let parts = split(input[1:], " ") | map(trim(#)) | filter(len(#) > 0);
  len(parts) == 1
    ? "Call the tool \"" + parts[0] + "\" with defaultPromptParameter as empty"
    : "Call the tool \"" + parts[0] + "\" with defaultPromptParameter as exactly \"" + join(parts[1:], " ") + "\""
```

- https://github.com/gptscript-ai/context/tree/main/available-agents

```go
Name: Available Agents
Description: List the agents that are available in the current context
Context: sys.context
Type: context

#!sys.interp.expr
// GPTSCRIPT_CONTEXT
let agents = GPTSCRIPT_CONTEXT.AgentGroup | filter(.ToolID != GPTSCRIPT_CONTEXT.CurrentAgent().ToolID) | map({"\t" + .Named + " " + GPTSCRIPT_CONTEXT.Program.ToolSet[.ToolID].Description});
len(agents) == 0 ? "" : "You have the following (" + string(len(agents)) + ") agent(s) available to you:\n" + join(agents, "\n")
```
I plan to check and try to convert more scripts and tools to get a better understanding, but please feel free to propose stuff that you would like me to check.

PR #830, opened 2024-08-28
interp(text/template): init text/template interpreter

rinor:
Was curious to check how it would look like using https://pkg.go.dev/text/template

Note: it might need a lot of helper functions to become useful

```go
Name: At syntax
Description: Add the ability for users to use at (@) to direct message an agent
Type: context

Share Input Filter: At syntax filter

---
Name: At syntax filter
Description: Translates @name command into "Call tool name with command"

#!sys.interp.template.text
{{ $input := .INPUT }}
{{ if or (eq $input nil) (eq (len $input) 0) }}
  {{ $input = "" }}
{{ else }}
  {{ $input = trimSpace $input }}
{{ end }}
{{ if or (lt (len $input) 2) (ne (hasPrefix $input "@") true) }}
{{ $input }}
{{ else }}
  {{ $input = trimSpace (trimPrefix $input "@") }}
  {{ $mparts := cut $input " " }}
  {{ $tool :=  trimSpace $mparts.before }}
  {{ $defaultPromptParameter :=  trimSpace $mparts.after }}
  {{ if eq (len $mparts.after) 0 }}
Call the tool "{{ $tool }}" with defaultPromptParameter as empty
  {{ else }}
Call the tool "{{ $tool }}" with defaultPromptParameter as exactly "{{ $defaultPromptParameter }}"
  {{ end }}
{{ end }}
```

```go
Name: Available Agents
Description: List the agents that are available in the current context
Context: sys.context
Type: context

#!sys.interp.template.text
{{ $currTID := .GPTSCRIPT_CONTEXT.Parent.CurrentAgent.ToolID }}
{{ $pts := .GPTSCRIPT_CONTEXT.Program.ToolSet }}
{{ $agents := newSlice }}
{{ $s := ""}}
{{ range .GPTSCRIPT_CONTEXT.AgentGroup }}
  {{ if (ne .ToolID $currTID)  }}
    {{ $agents = append $agents (print "\t" .Named " " (index $pts .ToolID).Description) }}
  {{ end }}
{{ end }}
{{ if (ne (len $agents) 0) }}
  {{ if (gt (len $agents) 1) }} {{ $s = "s" }} {{ end }}
  {{ print "You have the following " (len $agents) " agent" $s " available to you:\n" (join $agents "\n") }}
{{ end }}

```

### DIFF
--- a/pkg/builtin/builtin.go
+++ b/pkg/builtin/builtin.go
@@ -35,6 +35,7 @@ var SafeTools = map[string]struct{}{
 	"sys.time.now":                  {},
 	"sys.context":                   {},
 	"sys.model.provider.credential": {},
+	"sys.interp.template.text":      {},
 }
 
 var tools = map[string]types.Tool{
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -299,7 +299,10 @@ func (e *Engine) Start(ctx Context, input string) (ret *Return, _ error) {
 			return e.runOpenAPI(tool, input)
 		} else if tool.IsEcho() {
 			return e.runEcho(tool)
+		} else if tool.IsInterpTemplateText() {
+			return e.runInterpTemplateText(ctx, tool, input)
 		}
+
 		s, err := e.runCommand(ctx, tool, input, ctx.ToolCategory)
 		if err != nil {
 			return nil, err
--- a/pkg/engine/interpreter_template.go
+++ b/pkg/engine/interpreter_template.go
@@ -1,0 +1,100 @@
+package engine
+
+import (
+	"bytes"
+	"encoding/json"
+	"strings"
+	"text/template"
+
+	"github.com/gptscript-ai/gptscript/pkg/counter"
+	"github.com/gptscript-ai/gptscript/pkg/types"
+)
+
+func (e *Engine) runInterpTemplateText(ctx Context, tool types.Tool, input string) (*Return, error) {
+	// get code to eval
+	_, script, _ := strings.Cut(ctx.Tool.Instructions, "\n")
+	if script == "" {
+		return nil, nil
+	}
+
+	// setup environment for the interpreter
+	tmplData, err := prepareData(&ctx, input)
+	if err != nil {
+		return nil, err
+	}
+
+	tt, err := template.New(tool.Name).Funcs(prepareFuncMap()).Parse(script)
+	if err != nil {
+		return nil, err
+	}
+
+	// eval code
+	var tplOut bytes.Buffer
+	err = tt.Execute(&tplOut, tmplData)
+	if err != nil {
+		return nil, err
+	}
+	out := tplOut.String()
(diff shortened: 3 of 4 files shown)


ibuildthecloud:
I think this is reasonable. I'd probably call is `sys.template.golang`  Because supporting things like jinja or templ would be reasonable choices later.

rinor:
> FWIW I think golang templates are just about the worse syntax for users.

Indeed and I regretted as soon as I posted this. 
I'm looking for smth (**CGO free**) that's worth being embedded in gptscript, but this is not it. As an external interpreter may be feasible, but for embedded is not worth it atm.

> Then yaegi could also be an interpreter and it doesn't have to be a hard dep in gptscript.

Also yaegi embedded is not worth it atm, but anyone interested can already use it with default gptscript by (ab)using `#!/usr/bin/env yaegi` (assuming you already have [yaegi](https://github.com/traefik/yaegi/releases) binary). 

i.e both variants should work right now
- code in gpt file instructions
```go
Name: At syntax filter
Description: Translates @name command into "Call tool name with command"

#!/usr/bin/env yaegi
package main

func main() {
	....
}
```
- code in dedicated file

```go
Name: At syntax filter
Description: Translates @name command into "Call tool name with command"

#!/usr/bin/env yaegi /path_to/at-syntax.go
```

> I think what we really need is a way to make interpreters pluggable. So people can use whatever template language they want. For example.
> 
> ```
> #!sys.interpreter github.com/ibuildthecloud/golang-template
> 
> ....
> ```
> 
> So the syntax would be sys.interpeter followed by tool reference. Code wise this is ugly because it creates a circular reference from engine to runner, which could be cleanly done somehow I'm sure.
> 
> The mechanism would then be an interpreter will receive known args, similar to how input/output filters work. Not sure the exact args, but probably close to what you already have as the template context. If you could explore this route that would be great.

I agree and I'll certainly continue to investigate and explore this route while I check different solutions, but I guess this is not much of "blocker" since one, with enough interest, can already use `#!/usr/bin/env ...` with any cli interpreter (existing ones like with yaegi) or build one for the interested "language/template", maybe not ideal because it relies only on environment variables exposed by gptscript, but it should be fine for many tasks.

---
Just for Context:

The reason I'm looking for an embedded solution is because of gptscript deployments in isolated environments that can't use `fork/exec`, think unikernels,  so features that today rely on external applications, have to be provided by embedded interpreted environment (ofc agents/tools that depend on a binary like `gcloud`, `aws`, ... to function, will delegate the binary call to another location that can run that tool, think of webhook like server, by keeping a simple signature doable by `sys.http.post` and `openapi`). 
This is also the reason for #823 and #826.

---

For now I'll get back at #818 since to me it looks like the best candidate and it will help me investigate and learn more about gptscript internals and see what the plugable interface might look like.

rinor:
this is not worth embedding, closing it since I'll start working on a plugin based approach.

PR #832, opened 2024-08-29
enhance: avoid context limit

g-linville:
for gptscript-ai/desktop#283

This PR introduces three new measures to try to help the user avoid hitting the context length error:

1. Garbage collection works now. Previously, the token threshold was set too high, so we would run out of context before garbage collection would ever hit. Now, older messages will get deleted from the chat (on the LLM's side only; users can still see them) once it is needed in order to stay under the context limit.
2. If the output of a single tool output surpasses approximately 80% of the total context window, the tool output is considered too long and an error message is given to the LLM. We will design tools with the context limits in mind, but this is a safeguard to prevent just crashing and to allow the user to continue chatting with the LLM, and to allow the LLM to adjust its tool arguments and possibly remain under the output limit.
3. If, despite the previous measures, we do somehow hit the context length limit error, we catch that error and try to garbage collect more aggressively, and then retry the call up to ten times, doing more aggressive garbage collection each time.

With these three measures, I am fairly confident that users will no longer encounter the error.

### DIFF
--- a/pkg/openai/client.go
+++ b/pkg/openai/client.go
@@ -2,8 +2,10 @@ package openai
 
 import (
 	"context"
+	"errors"
 	"io"
 	"log/slog"
+	"math"
 	"os"
 	"slices"
 	"sort"
@@ -24,6 +26,7 @@ import (
 const (
 	DefaultModel    = openai.GPT4o
 	BuiltinCredName = "sys.openai"
+	TooLongMessage  = "Error: tool call output is too long"
 )
 
 var (
@@ -317,6 +320,14 @@ func (c *Client) Call(ctx context.Context, messageRequest types.CompletionReques
 	}
 
 	if messageRequest.Chat {
+		// Check the last message. If it is from a tool call, and if it takes up more than 80% of the budget on its own, reject it.
+		lastMessage := msgs[len(msgs)-1]
+		if lastMessage.Role == string(types.CompletionMessageRoleTypeTool) && countMessage(lastMessage) > int(math.Round(float64(getBudget(messageRequest.MaxTokens))*0.8)) {
+			// We need to update it in the msgs slice for right now and in the messageRequest for future calls.
+			msgs[len(msgs)-1].Content = TooLongMessage
+			messageRequest.Messages[len(messageRequest.Messages)-1].Content = types.Text(TooLongMessage)
+		}
+
 		msgs = dropMessagesOverCount(messageRequest.MaxTokens, msgs)
 	}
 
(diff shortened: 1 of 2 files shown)


ibuildthecloud (approved):
The very important thing is that we never drop the system role message. I believe that is still that case.

g-linville:
@ibuildthecloud correct. The way I implemented this will never change how the system role messages are handled. Those will always stick around.

pkg/openai/client.go
```diff
@@ -317,6 +320,14 @@ func (c *Client) Call(ctx context.Context, messageRequest types.CompletionReques
 	}
 
 	if messageRequest.Chat {
+		// Check the last message. If it is from a tool call, and if it takes up more than 80% of the budget on its own, reject it.
+		lastMessage := msgs[len(msgs)-1]
+		if lastMessage.Role == string(types.CompletionMessageRoleTypeTool) && countMessage(lastMessage) > int(math.Round(float64(getBudget(messageRequest.MaxTokens))*0.8)) {
```
thedadams, line 325: nit: probably don't need to `math.Round` here. Not much of a difference between 102,399 and 102,400.

pkg/openai/client.go
```diff
@@ -383,6 +394,16 @@ func (c *Client) Call(ctx context.Context, messageRequest types.CompletionReques
 		return nil, err
 	} else if !ok {
 		response, err = c.call(ctx, request, id, status)
+
+		// If we got back a context length exceeded error, keep retrying and shrinking the message history until we pass.
+		var apiError *openai.APIError
+		if err != nil && errors.As(err, &apiError) && apiError.Code == "context_length_exceeded" && messageRequest.Chat {
```
thedadams, line 400: nit: `errors.As` takes care of the `err != nil` check
```suggestion
		if errors.As(err, &apiError) && apiError.Code == "context_length_exceeded" && messageRequest.Chat {
```

pkg/openai/client.go
```diff
@@ -421,6 +442,32 @@ func (c *Client) Call(ctx context.Context, messageRequest types.CompletionReques
 	return &result, nil
 }
 
+func (c *Client) contextLimitRetryLoop(ctx context.Context, request openai.ChatCompletionRequest, id string, maxTokens int, status chan<- types.CompletionStatus) ([]openai.ChatCompletionStreamResponse, error) {
+	var (
+		response []openai.ChatCompletionStreamResponse
+		err      error
+	)
+
+	for range 10 { // maximum 10 tries
```
thedadams, line 451: Is this the first use in our code base?

pkg/openai/count.go
```diff
@@ -1,20 +1,32 @@
...
+	"math"
+
+	openai "github.com/gptscript-ai/chat-completion-client"
+)
+
+const DefaultMaxTokens = 128_000
+
+func decreaseTenPercent(maxTokens int) int {
+	maxTokens = getBudget(maxTokens)
+	return int(math.Round(float64(maxTokens) * 0.9))
```
thedadams, line 13: same nit about `math.Round`

thedadams (approved):

pkg/openai/client.go
g-linville: I think so!

pkg/openai/client.go
g-linville: Fixed

pkg/openai/client.go
g-linville: Fixed

pkg/openai/count.go
g-linville: Fixed

PR #837, opened 2024-09-03
fix: sys.read: never read files that contain a null byte

g-linville:
for gptscript-ai/desktop#263

### DIFF
--- a/pkg/builtin/builtin.go
+++ b/pkg/builtin/builtin.go
@@ -489,6 +489,12 @@ func SysRead(_ context.Context, _ []string, input string, _ chan<- string) (stri
 	if len(data) == 0 {
 		return fmt.Sprintf("The file %s has no contents", params.Filename), nil
 	}
+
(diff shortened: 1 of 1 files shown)


pkg/builtin/builtin.go
```diff
@@ -489,6 +489,12 @@ func SysRead(_ context.Context, _ []string, input string, _ chan<- string) (stri
 	if len(data) == 0 {
 		return fmt.Sprintf("The file %s has no contents", params.Filename), nil
 	}
+
+	// Assume the file is not text if it contains a null byte
+	if bytes.Contains(data, []byte{0}) {
```
thedadams, line 494: nit: `bytes.IndexByte` is probably slightly more performant.

thedadams (approved):

pkg/builtin/builtin.go
g-linville: Fixed

StrongMonkey (approved):

PR #838, opened 2024-09-05
fix: openapi revamp: improve operation not found messages

g-linville:


g-linville:
I totally re-did this PR because most of it was no longer necessary. What remains is the only helpful bit.

drpebcak (approved):

cjellick (approved):

PR #843, opened 2024-09-06
fix: openapi revamp: return error to LLM if args are invalid JSON

g-linville:


drpebcak (dismissed):

pkg/openapi/run.go
```diff
@@ -38,7 +38,8 @@ func Run(operationID, defaultHost, args string, t *openapi3.T, envs []string) (s
 	// Validate args against the schema.
 	validationResult, err := gojsonschema.Validate(gojsonschema.NewStringLoader(schemaJSON), gojsonschema.NewStringLoader(args))
 	if err != nil {
-		return "", false, err
+		// We don't return an error here because we want the LLM to be able to maintain control and try again.
+		return "ERROR: failed to validate arguments. Make sure your arguments are valid JSON.", false, nil
```
tylerslaton, line 42: Maybe its worth returning the actual error to the LLM as well so it knows what explicitly is wrong?

```suggestion
		return fmt.Sprintf("ERROR: failed to validate arguments. Make sure your arguments are valid JSON. %v", err), false, nil
```

tylerslaton (dismissed):
Had a suggestion but it isn't blocking for a merge.

drpebcak (approved):

cjellick (approved):

tylerslaton (approved):

PR #845, opened 2024-09-10
chore: bubble up errors when downloading credential helper fails

thedadams:


pkg/repos/runtimes/golang/golang.go
```diff
@@ -249,7 +252,8 @@ func (r *Runtime) Binary(ctx context.Context, tool types.Tool, _, toolSource str
 		return false, nil, nil
 	}
 
-	rel, ok := getLatestRelease(tool)
+	// ignore the error
```
StrongMonkey, line 255: Any reason we need to ignore the error here, while not on the `getLatestRelease` method?

thedadams: The `getLatestRelease` is best-effort because we don't know if the tool actually has a release. So errors don't really matter....

Except for the credential helpers. We know those have releases and we expect them to be used.

StrongMonkey (approved):

drpebcak (approved):

PR #846, opened 2024-09-12
feat: sdkserver: add credential routes

g-linville:
I've tested all of these with curl and they work properly.

### DIFF
--- a/pkg/sdkserver/credentials.go
+++ b/pkg/sdkserver/credentials.go
@@ -1,0 +1,176 @@
+package sdkserver
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	"github.com/gptscript-ai/gptscript/pkg/config"
+	gcontext "github.com/gptscript-ai/gptscript/pkg/context"
+	"github.com/gptscript-ai/gptscript/pkg/credentials"
+	"github.com/gptscript-ai/gptscript/pkg/repos/runtimes"
+)
+
+func (s *server) initializeCredentialStore(ctx string) (credentials.CredentialStore, error) {
+	cfg, err := config.ReadCLIConfig(s.gptscriptOpts.OpenAI.ConfigFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CLI config: %w", err)
+	}
+
+	// TODO - are we sure we want to always use runtimes.Default here?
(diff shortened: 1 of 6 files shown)


pkg/sdkserver/routes.go
```diff
@@ -58,6 +58,11 @@ func (s *server) addRoutes(mux *http.ServeMux) {
 	mux.HandleFunc("POST /confirm/{id}", s.confirm)
 	mux.HandleFunc("POST /prompt/{id}", s.prompt)
 	mux.HandleFunc("POST /prompt-response/{id}", s.promptResponse)
+
+	mux.HandleFunc("POST /credentials", s.listCredentials)
+	mux.HandleFunc("POST /credentials/create", s.createCredential)
+	mux.HandleFunc("POST /credentials/reveal", s.revealCredential)
```
g-linville, line 64: Let me know if these routes should change at all or if this is fine. Feels kinda weird to have it this way, but it's also abnormal to do a request body with a GET, so...I just did this.

pkg/sdkserver/credentials.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+func (s *server) initializeCredentialStore(ctx string) (credentials.CredentialStore, error) {
+	cfg, err := config.ReadCLIConfig(s.gptscriptOpts.OpenAI.ConfigFile)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read CLI config: %w", err)
+	}
+
+	// TODO - are we sure we want to always use runtimes.Default here?
+	store, err := credentials.NewStore(cfg, runtimes.Default(s.gptscriptOpts.Cache.CacheDir), ctx, s.gptscriptOpts.Cache.CacheDir)
```
g-linville, line 21: Right now, `runtimes.Default` is the only RuntimeManager in GPTScript, as far as I am aware. I would initialize this the normal way through `s.gptscriptOpts`, but the Runner opts there are nil, so I just did this instead.

pkg/sdkserver/routes.go
thedadams: The alternative is to use path and query parameters (e.g. `/credentials/reveal/name?context=my-context` or `/credentials/delete/name?context=my-context` (and make this `DELETE`)).

I leave the ultimate decision up to you. We certainly aren't building a REST API. This is more of RPC stuff, which I think is almost always `POST`.

I'll leave a final review after your response.

g-linville: Cool, I think I will leave it as it is then. All POST is what works best with the way the Go SDK is set up, at least.

thedadams (approved):

PR #848, opened 2024-09-13
fix: sdkserver: credentials: ensure credential helpers exist

g-linville:
The SDKServer was assuming that credential helpers were already present, which causes problems when they aren't there.

### DIFF
--- a/pkg/sdkserver/server.go
+++ b/pkg/sdkserver/server.go
@@ -17,6 +17,7 @@ import (
 	"github.com/gptscript-ai/broadcaster"
 	"github.com/gptscript-ai/gptscript/pkg/gptscript"
 	"github.com/gptscript-ai/gptscript/pkg/mvl"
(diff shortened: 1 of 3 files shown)


pkg/sdkserver/server.go
```diff
@@ -108,6 +109,7 @@ func run(ctx context.Context, listener net.Listener, opts Options) error {
 		token:            token,
 		client:           g,
 		events:           events,
+		runtimeManager:   runtimes.Default(opts.Options.Cache.CacheDir), // TODO - do we always want to use runtimes.Default here?
```
StrongMonkey, line 112: I like how you have moved the comment :)

StrongMonkey:
Always good to add a small description about the PR to have more contex

StrongMonkey (approved):
LGTM, please add some description to the PR 👍

drpebcak (approved):

PR #853, opened 2024-09-18
feat: add sqlite credential store

g-linville:
This adds a new credential store option, `sqlite`, which uses a pure Go SQLite driver with GORM to store credentials in a local SQLite database file.

By default, the SQLite file will be created at `<xdg config dir>/gptscript/credentials.db`, though this can be overridden with an environment variable. We also check for a Kubernetes `EncryptionConfiguration` file at `<xdg config dir>/gptscript/encryptionconfig.yaml` (this path can also be overridden with an environment variable). If this config file is present, we use it for configuration. We can do AES-GCM with a locally stored key, or something fancier like a KMS v2 plugin to support AWS KMS or some other third party service.

Here is an example encryptionconfig.yaml to use AES-GCM:

```yaml
kind: EncryptionConfiguration
apiVersion: apiserver.config.k8s.io/v1
resources:
- resources:
  - credentials
  providers:
  - aesgcm:
      keys:
      - name: myKey
        secret: <encryption key encoded in base64>
```

For now I think we want to leave this undocumented? Let me know if I should remove the references to it in the error output (the output we display to the user when they configure an invalid credential store in their config file).


go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/docker/docker-credential-helpers v0.8.1
 	github.com/fatih/color v1.17.0
 	github.com/getkin/kin-openapi v0.124.0
+	github.com/glebarez/sqlite v1.11.0
```
g-linville, line 14: This is the pure Go SQLite driver for GORM.

go.mod
```diff
@@ -35,7 +36,10 @@ require (
 	golang.org/x/sync v0.7.0
 	golang.org/x/term v0.22.0
 	gopkg.in/yaml.v3 v3.0.1
+	gorm.io/gorm v1.25.12
 	gotest.tools/v3 v3.5.1
+	k8s.io/apimachinery v0.31.1
+	k8s.io/apiserver v0.31.1
```
g-linville, line 42: I got approval from Darren to include whatever k8s dependencies were necessary to do encryption the same way we did it in mink.

g-linville:
Temporarily moved back to draft because I found a bug in my stacked credentials implementation. Working on it...

g-linville:
Should be good now.

g-linville:
Closing this since I am going to reimplement it as a separate application that we use as a credential helper.

PR #857, opened 2024-09-23
feat: support sqlite credential helper

g-linville:
This adds support for the new https://github.com/gptscript-ai/gptscript-credential-sqlite credential helper. It is for macOS and Linux. I will add some documentation about how to use it to its repo. By default, secrets will be unencrypted in the database, but we support using k8s EncryptionConfiguration files, which can be used to set up AES encryption, or something more sophisticated with a KMS like AWS KMS.

### DIFF
--- a/pkg/config/cliconfig.go
+++ b/pkg/config/cliconfig.go
@@ -16,21 +16,31 @@ import (
 )
 
 const (
-	Wincred       = "wincred"
-	Osxkeychain   = "osxkeychain"
-	Secretservice = "secretservice"
-	Pass          = "pass"
-	File          = "file"
-	Sqlite        = "sqlite"
+	WincredCredHelper       = "wincred"
+	OsxkeychainCredHelper   = "osxkeychain"
+	SecretserviceCredHelper = "secretservice"
+	PassCredHelper          = "pass"
+	FileCredHelper          = "file"
+	SqliteCredHelper        = "sqlite"
 
 	GPTScriptHelperPrefix = "gptscript-credential-"
 )
 
 var (
-	darwinHelpers  = []string{Osxkeychain, File, Sqlite}
-	windowsHelpers = []string{Wincred, File}
-	linuxHelpers   = []string{Secretservice, Pass, File, Sqlite}
+	darwinHelpers  = []string{OsxkeychainCredHelper, FileCredHelper, SqliteCredHelper}
+	windowsHelpers = []string{WincredCredHelper, FileCredHelper}
+	linuxHelpers   = []string{SecretserviceCredHelper, PassCredHelper, FileCredHelper, SqliteCredHelper}
 )
+
+func listAsString(helpers []string) string {
+	if len(helpers) == 0 {
+		return ""
+	} else if len(helpers) == 1 {
+		return helpers[0]
+	}
+
+	return strings.Join(helpers[:len(helpers)-1], ", ") + " or " + helpers[len(helpers)-1]
+}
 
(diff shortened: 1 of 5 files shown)


cjellick:
test from gptscript

pkg/config/cliconfig.go
```diff
@@ -15,13 +15,22 @@ import (
...
-	darwinHelpers  = []string{"osxkeychain", "file"}
-	windowsHelpers = []string{"wincred", "file"}
-	linuxHelpers   = []string{"secretservice", "pass", "file"}
+const (
+	Wincred       = "wincred"
+	Osxkeychain   = "osxkeychain"
+	Secretservice = "secretservice"
+	Pass          = "pass"
+	File          = "file"
+	Sqlite        = "sqlite"
```
thedadams, line 24: I think appending `CredHelper` to each of these would give them a more descriptive name.

pkg/config/cliconfig.go
```diff
@@ -150,11 +159,11 @@ func ReadCLIConfig(gptscriptConfigFile string) (*CLIConfig, error) {
 		errMsg := fmt.Sprintf("invalid credential store '%s'", result.CredentialsStore)
 		switch runtime.GOOS {
 		case "darwin":
-			errMsg += " (use 'osxkeychain' or 'file')"
+			errMsg += " (use 'osxkeychain', 'file', or 'sqlite')"
```
thedadams, line 162: These probably won't change much, but you could use `strings.Join(darwinHelpers, ", ")` here for simplicity.

pkg/credentials/store.go
```diff
@@ -178,7 +178,7 @@ func (s *Store) getStore(ctx context.Context) (credentials.Store, error) {
 }
 
 func (s *Store) getStoreByHelper(ctx context.Context, helper string) (credentials.Store, error) {
-	if helper == "" || helper == config.GPTScriptHelperPrefix+"file" {
+	if helper == "" || helper == config.GPTScriptHelperPrefix+config.File {
```
thedadams, line 181: This solidifies my desire to change the names of those constants: `config.File` is a confusing name.

pkg/config/cliconfig.go
g-linville: Fixed

pkg/config/cliconfig.go
g-linville: Fixed

pkg/credentials/store.go
g-linville: :+1:

thedadams (approved):

pkg/credentials/util.go
```diff
@@ -1,18 +1,43 @@
...
 
 type CredentialHelperDirs struct {
 	RevisionFile, LastCheckedFile, BinDir string
 }
 
-func GetCredentialHelperDirs(cacheDir string) CredentialHelperDirs {
+func RepoNameForCredentialStore(store string) string {
+	switch store {
+	case config.SqliteCredHelper:
+		return "gptscript-credential-sqlite"
```
StrongMonkey, line 18: Nit: `gptscript-credential-sqlite` and `gptscript-credential-helpers` shows up multiple and might be a good idea to define those as const.

StrongMonkey (approved):

pkg/credentials/util.go
g-linville: I might change this at a later point. Don't want to go through the process of getting approvals on this one again just for that change

PR #869, opened 2024-10-10
feat: add dataset operations to sdkserver

g-linville:
This adds five new operations to the SDKServer to run the tools that we are using as the API for datasets. The SDKServer is basically just acting as a middleman between the client using the SDK, and the dataset API (the five tools in https://github.com/gptscript-ai/datasets/pull/1).

### DIFF
--- a/pkg/gptscript/gptscript.go
+++ b/pkg/gptscript/gptscript.go
@@ -32,8 +32,6 @@ import (
 
 var log = mvl.Package()
 
-const defaultDatasetToolRepo = "github.com/gptscript-ai/datasets"
-
 type GPTScript struct {
 	Registry               *llm.Registry
 	Runner                 *runner.Runner
@@ -83,9 +81,6 @@ func Complete(opts ...Options) Options {
 	}
 	if len(result.CredentialContexts) == 0 {
 		result.CredentialContexts = []string{credentials.DefaultCredentialContext}
-	}
-	if result.DatasetToolRepo == "" {
-		result.DatasetToolRepo = defaultDatasetToolRepo
 	}
 
 	return result
--- a/pkg/gptscript/gptscript.go
+++ b/pkg/gptscript/gptscript.go
@@ -32,6 +32,8 @@ import (
 
 var log = mvl.Package()
 
+const defaultDatasetToolRepo = "github.com/gptscript-ai/datasets"
+
 type GPTScript struct {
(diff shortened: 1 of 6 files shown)


pkg/gptscript/gptscript.go
```diff
@@ -80,6 +84,9 @@ func Complete(opts ...Options) Options {
 	if len(result.CredentialContexts) == 0 {
 		result.CredentialContexts = []string{credentials.DefaultCredentialContext}
 	}
+	if result.DatasetToolRepo == "" {
+		result.DatasetToolRepo = defaultDatasetToolRepo
+	}
```
thedadams, line 89: I don't think this is needed given that you have a default set.

I do believe it is needed here, though, when running the SDK server in embedded mode: https://github.com/gptscript-ai/gptscript/blob/45d444f810600d9584951ec3aba98da9b5e02db5/pkg/sdkserver/server.go#L154

g-linville: Thanks. This is fixed

pkg/sdkserver/datasets.go
```diff
@@ -0,0 +1,332 @@
...
+		writeError(logger, w, http.StatusInternalServerError, fmt.Errorf("failed to run program: %w", err))
+		return
+	}
+
+	writeResponse(logger, w, map[string]any{"stdout": result})
+}
+
+type createDatasetArgs struct {
+	Name        string `json:"dataset_name"`
+	Description string `json:"dataset_description"`
```
thedadams, line 85: nit: you have camelCase and snake_case in your `json` tags. Can this be standardized?

thedadams (approved):

pkg/sdkserver/datasets.go
g-linville: fixed

thedadams: This went the opposite way I expected. Is there reason why you are doing snake_case? I think most of our stuff uses camelCase.

g-linville: Lol. I honestly have no idea. I'll go change it all to camelCase.

g-linville: Fixed. Sorry for the delay. Got stuck working on a test for another repo

thedadams (approved):

njhale (approved):

PR #872, opened 2024-10-14
feat: add workspace API to SDK

thedadams:
This change also changes the error behavior when running tools that are simply wrapped commands. Previously, all such tools would not return an error, rather an error message in hopes that the LLM would retry. However, if the tool is just a command (i.e. has no parent), then it should return an error so that the caller doesn't have to guess whether an error occurred.

### DIFF
--- a/pkg/sdkserver/workspaces.go
+++ b/pkg/sdkserver/workspaces.go
@@ -1,0 +1,328 @@
+package sdkserver
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	gcontext "github.com/gptscript-ai/gptscript/pkg/context"
+	"github.com/gptscript-ai/gptscript/pkg/loader"
+)
+
+type workspaceCommonRequest struct {
+	ID                string `json:"id"`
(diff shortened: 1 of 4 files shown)


pkg/sdkserver/workspaces.go
```diff
@@ -0,0 +1,328 @@
...
+type workspaceCommonRequest struct {
+	ID                string `json:"id"`
+	WorkspaceToolRepo string `json:"workspaceToolRepo"`
+}
+
+func (w workspaceCommonRequest) getToolRepo() string {
+	if w.WorkspaceToolRepo != "" {
+		return w.WorkspaceToolRepo
+	}
+	return "/Users/thedadams/code/workspace-provider"
```
g-linville, line 21: Should be updated before merging

g-linville (approved):

pkg/sdkserver/workspaces.go
thedadams: Thanks. I thought I got this, but, alas, I forgot.

g-linville (approved):

PR #908, opened 2024-11-18
fix: credentials: only decrypt credentials in the context(s) needed

g-linville:


pkg/credentials/toolstore.go
```diff
@@ -80,16 +80,10 @@ func (h *toolCredentialStore) GetAll() (map[string]types.AuthConfig, error) {
...
-
-	for serverAddress := range newCredAddresses {
-		ac, err := h.Get(serverAddress)
-		if err != nil {
-			return nil, err
+		result[toolNameWithCtx(toolName, ctx)] = types.AuthConfig{
+			Username:      val,
+			ServerAddress: serverAddress,
 		}
-		result[serverAddress] = ac
```
g-linville, line 92: This loop is the main thing I needed to fix. Calling `h.Get` results in the credential getting possibly decrypted. I refactored this so that we no longer call `h.Get` here in the GetAll function, and instead call it in the other file, and only on the credentials that we are actually going to return.

PR #909, opened 2024-11-21
Fix server creds

thedadams:
This change does two related things:
1. Allow listing of credentials in all contexts
2. When running in SDK server mode and using the file cred store, we must 
reload the credentials in the file each time. Typically, these 
credentials will be added via a tool call, which would invalidate the 
credentials held in memory.

### DIFF
--- a/pkg/config/cliconfig.go
+++ b/pkg/config/cliconfig.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/adrg/xdg"
 	"github.com/docker/cli/cli/config/types"
+	"github.com/gptscript-ai/gptscript/pkg/mvl"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/config/cliconfig.go
```diff
@@ -116,13 +118,21 @@ func (c *CLIConfig) GetAuthConfigs() map[string]types.AuthConfig {
...
-		c.auths = map[string]types.AuthConfig{}
-		for k, v := range c.Auths {
-			authConfig := (types.AuthConfig)(v)
-			c.auths[k] = authConfig
-		}
+		c.auths = make(map[string]types.AuthConfig, len(c.Auths))
+	}
+
+	// Assume that whatever was pulled from the file is more recent.
+	// The docker creds framework with save the file after creating or updating a credential.
```
iwilltry42, line 131: Super nit:
```suggestion
	// The docker creds framework will save the file after creating or updating a credential.
```

iwilltry42 (dismissed):

StrongMonkey (dismissed):

njhale (approved):

PR #915, opened 2024-12-04
enhance: add mTLS between gptscript and daemon tools

g-linville:


g-linville:
alternative model providers will likely continue to fail in the tests until they are updated for mTLS

PR #921, opened 2024-12-20
chore: add auth provider credential type

g-linville:
This is needed for the auth provider refactoring that I am doing.


g-linville:
Talked with Donnie, and we are just going to use the `tool` credential type for this, since that should work without any issues. Closing

PR #924, opened 2025-01-02
enhance: add a maximum consecutive tools calls restriction

thedadams:
This is a safety net for instances when the LLM goes awry. There is an undocumented environment variable to change the maximum number of consecutive tool calls from the default of 10.

Issue: https://github.com/obot-platform/obot/issues/940

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"slices"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


g-linville (approved):

pkg/engine/engine.go
```diff
@@ -385,6 +398,28 @@ func (e *Engine) complete(ctx context.Context, state *State) (*Return, error) {
 		}
 	}()
 
+	// Limit the number of consecutive tool calls and responses.
+	// We don't want the LLM to call tools unrestricted or get stuck in an error loop.
+	var messagesSinceLastUserMessage int
+	for _, msg := range slices.Backward(state.Completion.Messages) {
```
njhale, line 404: Using the new iterator functions, nice! 🎉

njhale (approved):

PR #932, opened 2025-01-14
feat: add file revision API for workspaces

thedadams:


iwilltry42 (approved):
Having those "descriptive Names" in the code to reference sub-tools somehow still scares me :grimacing:

PR #937, opened 2025-01-22
Chore: Support openai o1 model

StrongMonkey:
Openai O1 model doesn't support stream, and doesn't support setting temperature on chat completion. We have to tweak that in order to support o1 model.

https://github.com/obot-platform/obot/issues/1131

### DIFF
--- a/pkg/openai/client.go
+++ b/pkg/openai/client.go
@@ -545,7 +545,7 @@ func override(left, right string) string {
 	return left
 }
 
-func (c *Client) call(ctx context.Context, request openai.ChatCompletionRequest, transactionID string, env []string, partial chan<- types.CompletionStatus) (types.CompletionMessage, error) {
+func (c *Client) call(ctx context.Context, request openai.ChatCompletionRequest, transactionID string, envs []string, partial chan<- types.CompletionStatus) (types.CompletionMessage, error) {
 	streamResponse := os.Getenv("GPTSCRIPT_INTERNAL_OPENAI_STREAMING") != "false"
 
 	partial <- types.CompletionStatus{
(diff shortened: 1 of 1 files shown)


pkg/openai/client.go
```diff
@@ -259,8 +259,12 @@ func toMessages(request types.CompletionRequest, compat bool) (result []openai.C
 	}
 
 	if len(systemPrompts) > 0 {
+		role := types.CompletionMessageRoleTypeSystem
+		if useO1Model {
```
StrongMonkey, line 263: According to docs, for o1 model it is better to use developer message.

pkg/openai/client.go
```diff
@@ -446,6 +455,22 @@ func (c *Client) Call(ctx context.Context, messageRequest types.CompletionReques
 	return &result, nil
 }
 
+func isO1Model(model string, envs []string) bool {
```
StrongMonkey, line 458: there is two ways to check whether the model is used by o1

1. check if modeName is o1. This is true if used with standalone gptscript.
2. Check if `OPENAI_MODEL_NAME` is set. This will be set in Obot to determine the name.

StrongMonkey:
I changed the approach and move the logic to openai-model-provider. However, we still need a way to dynamically turn off stream since o1 doesn't support it. So the way to do that is to check `GPTSCRIPT_INTERNAL_OPENAI_STREAMING` exists in envs so that we could dynamic set it in obot.

In plain gptscript user would be expected to set that when they are using o1 model, as there is no smart way to automatically do this though.

StrongMonkey:
Closing in favor of adding the support from openai-proxy.

PR #942, opened 2025-01-30
fix: count tokens from tool definitions when adjusting for context window

g-linville:
The tokens used for tool definitions are counted as part of the context window, but we were failing to count them here when trying to stay beneath the limit. This introduces a fix so that we account for it.

Our token estimation still isn't great. For example, a message consisting of `1\n` repeated 30000 times is counted as 20000 tokens in our method (since we just divide the character count by 3), but in OpenAI's tokenizer, it is 60000 tokens, meaning we grossly underestimate. I think we should consider calling out to Python to use the `tiktoken` library to count tokens for us, as that will be far more reliable. However, in practice, this rarely seems to cause us any trouble, so maybe we are fine sticking with our current approach of just dividing by 3.


g-linville:
There is more work that would be needed here and it's just not a priority right now.

PR #943, opened 2025-02-03
fix: run daemon tools as tools when referenced by other tools

thedadams:
Daemon tools are usually started when other tools reference them by URL. However, they are not run like other tools. Therefore, their credentials are not processed as expected.

This change will detect when tools reference daemon tools and ensure they are run like other tools.


iwilltry42 (approved):

g-linville (approved):

thedadams:
Closing after feedback.

PR #951, opened 2025-03-11
enhance: add method for recreating all credentials

g-linville:
for obot-platform/obot#1997

### DIFF
--- a/pkg/credentials/factory.go
+++ b/pkg/credentials/factory.go
@@ -72,8 +72,8 @@ func (s *StoreFactory) NewStore(credCtxs []string) (CredentialStore, error) {
 		return nil, err
 	}
 	if s.file {
-		return withOverride{
-			target: Store{
+		return &withOverride{
+			target: &Store{
 				credCtxs: credCtxs,
 				cfg:      s.cfg,
 			},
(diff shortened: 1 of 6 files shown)


drpebcak (approved):

pkg/credentials/factory.go
```diff
@@ -72,17 +72,17 @@ func (s *StoreFactory) NewStore(credCtxs []string) (CredentialStore, error) {
 		return nil, err
 	}
 	if s.file {
-		return withOverride{
-			target: Store{
+		return &withOverride{
```
thedadams, line 75: Do I understand correctly that this change was made to avoid copying the mutex?

pkg/credentials/store.go
```diff
@@ -199,6 +217,54 @@ func (s Store) List(_ context.Context) ([]Credential, error) {
...
+		}
+
+		if err := store.Store(authConfig); err != nil {
+			s.recreateAllLock.Unlock()
+			return err
+		}
+		s.recreateAllLock.Unlock()
+	}
+
+	return nil
```
thedadams, line 265: Writing a helper function for this will allow you to use `defer s.recreateAllLock.Unlock()`.

thedadams (changes requested):

njhale (dismissed):
lgtm once Donnie's feedback is addressed.

pkg/credentials/factory.go
g-linville: Yep

pkg/credentials/store.go
g-linville: oh good call

g-linville: fixed

thedadams (approved):

PR #959, opened 2025-04-09
WIP

g-linville:


pkg/openai/count.go
```diff
@@ -54,22 +67,44 @@ func dropMessagesOverCount(maxTokens int, msgs []openai.ChatCompletionMessage) (
...
-	return append(result, msgs[withinBudget:]...)
+	return append(result, msgs[withinBudget:]...), nil
 }
 
-func countMessage(msg openai.ChatCompletionMessage) (count int) {
-	count += len(msg.Role)
-	count += len(msg.Content)
+func countMessage(msg openai.ChatCompletionMessage) (int, error) {
+	tiktoken.SetBpeLoader(tiktoken_loader.NewOfflineLoader())
+	encoding, err := tiktoken.GetEncoding("o200k_base")
```
njhale, line 78: can we set the loader once on package init?

njhale (approved):
one question, but otherwise lgtm

pkg/openai/count.go
g-linville: Yep, good call. Just pushed the change for this

njhale (approved):

thedadams (approved):

PR #973, opened 2025-05-20
chore: switch to huma for OpenAPI v3.1 support

thedadams:


pkg/loader/openapi.go
```diff
@@ -373,6 +373,147 @@ func parseServer(server *openapi3.Server) (string, error) {
 	return s, nil
 }
 
+// openAPI3SchemaToHumaV2Schema converts an openapi3.Schema to a humav2.Schema
+func openAPI3SchemaToHumaV2Schema(schema *openapi3.Schema) *humav2.Schema {
```
thedadams, line 377: Junie did this for me. I checked it out and it seems good, but would appreciate some eyes on this.

pkg/loader/openapi.go
```diff
@@ -373,6 +373,147 @@ func parseServer(server *openapi3.Server) (string, error) {
...
+	if schema.Max != nil {
+		maxVal := *schema.Max
+		result.Maximum = &maxVal
+
+		// In OpenAPI 3, ExclusiveMax is a boolean flag that applies to Max
+		// In Huma, ExclusiveMaximum is a separate value
+		if schema.ExclusiveMax {
+			result.ExclusiveMaximum = &maxVal
+		}
+	}
```
thedadams, line 419: This is the real reason we needed to switch.

pkg/loader/openapi.go
g-linville: image

g-linville (approved):

njhale (approved):

PR #977, opened 2025-05-30
fix: distinguish between nil and non-nil allowed tools

thedadams:
If allowedTools is nil, then all tools are allowed. If allowed tools is non-nil and has length zero, then no tools are allowed.


thedadams:
Is this a good idea?

g-linville:
I think this makes sense.

PR #979, opened 2025-06-11
enhance: provide contexts to credential helpers when listing credentials

g-linville:
for https://github.com/obot-platform/obot/issues/3180

This PR changes the way we request a list of credentials by also including the contexts as input. The contexts are ignored by all non-database credential helpers (like Wincred and macOS Keychain), so we still filter by context after we call `GetAll`, but this allows the database credential helpers to be more efficient and only return to us the credentials that we actually need.

Once this PR is merged, I will have a PR for Obot to bump the GPTScript dependency. We might have to temporarily (or permanently, as I don't think it will do any harm) enable schema migration for the credentials table, which is currently disabled with an environment variable.

Related PR to change the credential helpers: https://github.com/obot-platform/tools/pull/692

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,7 @@ init-docs:
 
 # Ensure docs build without errors. Makes sure generated docs are in-sync with CLI.
 validate-docs: gen-docs
-	docker run --rm --workdir=/docs -v $${PWD}/docs:/docs node:18-buster yarn build
+	docker run --rm --workdir=/docs -v $${PWD}/docs:/docs node:18-buster npm run build
 	if [ -n "$$(git status --porcelain --untracked-files=no)" ]; then \
 		git status --porcelain --untracked-files=no; \
 		echo "Encountered dirty repo!"; \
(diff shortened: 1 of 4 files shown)


StrongMonkey (approved):

njhale (approved):

g-linville:
Re-requesting review because I added a test.

Makefile
```diff
@@ -53,7 +53,7 @@ init-docs:
 
 # Ensure docs build without errors. Makes sure generated docs are in-sync with CLI.
 validate-docs: gen-docs
-	docker run --rm --workdir=/docs -v $${PWD}/docs:/docs node:18-buster yarn build
+	docker run --rm --workdir=/docs -v $${PWD}/docs:/docs node:18-buster npm run build
```
g-linville, line 56: I did this for consistency with the other change from yarn to npm in the `init-docs` target. Plus it wasn't working with Node v18 and yarn.

njhale (approved):

PR #980, opened 2025-06-12
chore: replace mcp-go with nanobot

thedadams:


Makefile
```diff
@@ -49,7 +49,7 @@ serve-docs:
 
 # This will initialize the node_modules needed to run the docs dev server. Run this before running serve-docs
 init-docs:
-	docker run --rm --workdir=/docs -v $${PWD}/docs:/docs node:18-buster yarn install
+	docker run --rm --workdir=/docs -v $${PWD}/docs:/docs node:18-buster npm install
```
thedadams, line 52: Is this OK? It looked like we are using npm everywhere else.

thedadams:
Looks like there's an issue here. Moving to draft while I figure it out.

g-linville (dismissed):
Nice, this is a lot cleaner

PR #987, opened 2025-06-23
test: remove mistral from smoke tests

njhale:
The smoke tests for `mistral-large-2402` are failing because its responses for certain test cases have changed since the initial golden files were generated. After a bit of investigation, it looks like this model was deprecated in November 2024 and retired in June 2025. Also, from what I can tell, contemporary Mistral models like `mistral-large-2411` and `mistral-medium-2505` don't actually make the expected tool calls required to pass the GPTScript's smoke tests, so swapping out the model is not feasible at this time.

In light of these findings, remove mistral from the set of smoke tested model providers.


njhale:
Note: The mistral smoke tests will still run on this PR since the current workflow file from main is used for PR runs.

thedadams (approved):

njhale:
Going to go ahead and merge this with @thedadams review.

PR #1016, opened 2026-04-09
chore: bump go modules

njhale:
Bump go modules to pick up CVE fixes.

### DIFF
--- a/.github/workflows/integration.yaml
+++ b/.github/workflows/integration.yaml
@@ -29,7 +29,7 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           cache: false
-          go-version: "1.23"
+          go-version: "1.26"
       - name: Build
         if: matrix.os == 'ubuntu-22.04'
         run: make build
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -22,7 +22,7 @@ smoke: build
 smoke:
 	go test -v -tags='smoke' ./pkg/tests/smoke/...
 
-GOLANGCI_LINT_VERSION ?= v2.6.1
+GOLANGCI_LINT_VERSION ?= v2.9.0
```
njhale, line 25: Bumping this to align with the version in Obot, since I was getting errors validating locally by mismatched versions.

.github/workflows/integration.yaml
```diff
@@ -29,7 +29,7 @@ jobs:
       - uses: actions/setup-go@v5
         with:
           cache: false
-          go-version: "1.23"
+          go-version: "1.26"
```
njhale, line 32: figured I'd update everything to 1.26 while I'm here.

pkg/context/context.go
```diff
@@ -1,4 +1,4 @@
-package context
+package context //nolint:revive
```
njhale, line 1: Needed to add these when I bumped golangci-lint to prevent new validation failures.

g-linville (approved):

